Issue #12, opened 2024-07-26
Feat: Propose .repopackignore file to override .gitignore for repopack bundles


MassDo:
Hello,

First, thank you for this very useful project!

I'd like to suggest a feature that could enhance repopack's flexibility. 

### Current situation:
I have a file "claude-custom-instructions.txt" that I want to include in my repopack bundle, but it's listed in my .gitignore file. Currently, to include this file, I need to set `ignore.useDefaultPatterns = false` and then manually add all the patterns I want to ignore in `ignore.customPatterns`. This works, but it's a bit cumbersome.

### Feature proposal:
Introduce a new file called `.repopackignore` that would override .gitignore settings specifically for repopack bundles when present. Here's how it could work:

1. If a `.repopackignore` file exists in the repository, repopack would use it instead of .gitignore.
2. The syntax would be similar to .gitignore, allowing users to easily specify which files to include or exclude from the bundle.

Benefits:
- Simpler management of bundle-specific ignores
- Maintains separation between git ignore rules and repopack bundle rules
- Allows users to easily include files in repopack bundles that are git-ignored

I'd appreciate your thoughts on this idea and whether it aligns with the project's goals. 

Thank you again !

yamadashy:
Hello,

Thank you for your valuable feature proposal!
You're absolutely right; the current setup does cause the issue. Your idea is great and `.repopackignore` would not only allow including git-ignored files but also provide an easier way to specify ignore patterns compared to the current `ignore.customPatterns` configuration.

However, I've been considering a slightly different approach that might offer even more flexibility. Here's what I'm thinking:

1. Introduce a new `.repopackignore` file for Repopack-specific ignore patterns, as you suggested.

2. Add a new configuration option `useGitignorePatterns` (default: true) in `repopack.config.json`:
   ```json5
   {
     "ignore": {
       "useGitignorePatterns": true,
       // other existing options...
     }
   }
   ```

Key points:
- If set to `false` and use `.repopackignore`, it would solve your issue by allowing Repopack to include git-ignored files.
- If set to `true` (default), it would merge `.gitignore` and `.repopackignore` patterns. This addresses a potential future requirement I've envisioned.

This approach would maintain flexibility for both current needs and potential future requirements.

What do you think about this alternative? Does it still address your needs? I'm very open to your feedback and any further suggestions you might have.

Thank you again for helping improve Repopack!

yamadashy:
Hello,
[...]

PR #13, opened 2024-07-27
Add support for .repopackignore and enhance ignore pattern documentation

yamadashy:
fix: #12 
This pull request introduces support for .repopackignore files and improves the documentation for ignore patterns in Repopack.

Key changes:
- Added support for .repopackignore files
- Updated the "Ignore Patterns" section in the README
- Rearrange priority order for ignore patterns

Ignore priority order (from highest to lowest):
1. Custom patterns `ignore.customPatterns`
2. `.repopackignore`
3. `.gitignore` (if `ignore.useGitignorePatterns` is true)
4. Default patterns (if `ignore.useDefaultPatterns` is true)


MassDo:
Hello,

Thank you for your quick and thoughtful response! Your proposed approach sounds perfect and addresses the issue comprehensively.

I do have one small suggestion regarding the naming of the configuration option. Instead of **`useGitignorePatterns`**, perhaps **`useGitignore`** or **`useGitignoreFile`** would be clearer. This can help avoid potential confusion with a setting that might suggest switching from a wildcard format to a regex format.

Best regards,

yamadashy:
Thank you for taking a look and for your suggestion! I agree that your proposed option is clearer.
I think we'll go with the simple useGitignore option.

yamadashy:
I've implemented the changes. Your input has helped improve the feature!
I'm going to merge this pull request now. Thanks again for your contribution to making Repopack better!

PR #113, opened 2024-10-09
Add feature split files by maxTokenSize per file

fridaystreet:


src/config/configLoad.ts
```diff
@@ -85,7 +85,7 @@ export const mergeConfigs = (
   // If the output file path is not provided in the config file or CLI, use the default file path for the style
   if (cliConfig.output?.filePath == null && fileConfig.output?.filePath == null) {
     const style = cliConfig.output?.style || fileConfig.output?.style || defaultConfig.output.style;
-    defaultConfig.output.filePath = defaultFilePathMap[style];
+    defaultConfig.output.filePath = defaultFilePathMap[style ?? 'plain'];
```
yamadashy, line 88: This change is not necessary since the defaultConfig will eventually make it plain.
```suggestion
    defaultConfig.output.filePath = defaultFilePathMap[style];
```

src/cli/cliRun.ts
```diff
@@ -44,6 +45,7 @@ export async function run() {
       .option('--init', 'initialize a new repopack.config.json file')
       .option('--global', 'use global configuration (only applicable with --init)')
       .option('--remote <url>', 'process a remote Git repository')
+      .option('--max-tokens <number>', 'maximum number of tokens per output file', Number.parseInt) // Add the maxTokens option
```
yamadashy, line 48: I would like to make it output-max-tokens if possible since it is max-tokens regarding output. I am concerned about the possibility of max-tokens for other uses in the future.
```suggestion
      .option('--output-max-tokens <number>', 'maximum number of tokens per output file', Number.parseInt) // Add the maxTokens option
```

src/cli/actions/remoteAction.ts
```diff
@@ -29,7 +29,7 @@ export const runRemoteAction = async (repoUrl: string, options: CliOptions): Pro
     logger.log('');
 
     const result = await runDefaultAction(tempDir, tempDir, options);
-    await copyOutputToCurrentDirectory(tempDir, process.cwd(), result.config.output.filePath);
+    await copyOutputToCurrentDirectory(tempDir, process.cwd(), result.config.output.filePath ?? 'repopack-output.txt');
```
yamadashy, line 32: It is not necessary because it will eventually become the value of defaultConfig in the process of config merging.
```suggestion
    await copyOutputToCurrentDirectory(tempDir, process.cwd(), result.config.output.filePath);
```

src/config/configTypes.ts
```diff
@@ -23,16 +27,7 @@ interface RepopackConfigBase {
...
-    filePath: string;
-    style: RepopackOutputStyle;
-    headerText?: string;
-    instructionFilePath?: string;
-    removeComments: boolean;
-    removeEmptyLines: boolean;
-    topFilesLength: number;
-    showLineNumbers: boolean;
-  };
+  output: RepopackOutputConfig;
```
yamadashy, line 30: Since defaultConfig dares to make some parts of the code non-nullable, I would like it to be the original code with this option added.

src/core/output/outputStyles/xmlStyle.ts
```diff
@@ -75,7 +85,7 @@ This section contains a summary of this file.
 <repository_files>
 This section contains the contents of the repository's files.
 
-{{#each processedFiles}}
+{{#each includedFiles}}
```
yamadashy, line 88: Please make the same correction to plainStyle.ts and markdownStyle.ts.

src/cli/actions/defaultAction.ts
```diff
@@ -56,8 +56,8 @@ export const runDefaultAction = async (
   spinner.succeed('Packing completed successfully!');
```
yamadashy, line 56: Regarding the whole thing,
please fix the tests that fail on npm run lint run and npm run test

src/core/output/outputStyles/plainStyle.ts
```diff
@@ -59,6 +63,11 @@ Usage Guidelines:
 -----------------
 {{{summaryUsageGuidelines}}}
 
+Repository Size:
+-----------------
```
yamadashy, line 67: This hyphen is aligned with the letter above.
```suggestion
----------------
```

src/core/output/outputSplitter.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const splitOutput = (
+  processedFiles: ProcessedFile[],
+  maxTokensPerFile: number,
+): OutputSplit[] => {
+  const tokenCounter = new TokenCounter();
+  const outputSplits: OutputSplit[] = [];
+  let currentTokenCount = 0;
+  let currentOutput = '';
```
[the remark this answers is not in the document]
yamadashy: Please correct this one.

src/core/output/outputSplitter.ts
```diff
@@ -0,0 +1,53 @@
...
+  let currentIncludedFiles: ProcessedFile[] = []; // Initialize currentIncludedFiles
+
+  for (const file of processedFiles) {
+    const fileTokenCount = tokenCounter.countTokens(file.content, file.path);
```
yamadashy, line 21: If there is a large number of files, the heavy processing of token count may be affected. Currently, the parts where token counts are being performed are slightly mitigated by adding sleep.

However, I thought it would be better to leave it as it is for now and fix it when the problem actually occurs.

src/core/packager.ts
```diff
@@ -72,13 +78,37 @@ export const pack = async (
...
-  // Write output to file. path is relative to the cwd
-  progressCallback('Writing output file...');
-  const outputPath = path.resolve(config.cwd, config.output.filePath);
-  logger.trace(`Writing output to: ${outputPath}`);
-  await fs.writeFile(outputPath, output);
+  // Write output to file(s). path is relative to the cwd
+  progressCallback('Writing output file(s)...');
+
+  // Handle the case where filePath is undefined
+  if (config.output.filePath) {
```
yamadashy, line 92: The filePath is set in the default config, so it will always be present.

kagux:
@fridaystreet appreciate a lot that you work on this feature! I've tried using the build from this PR and it produces empty files. They only include the header section with directory/files and nothing from files themselves. Even if no max token option is specified, still single file with no contents besides header.

fridaystreet:
@kagux thanks for the feedback. It was all working and seems to be wirking locally fir me. But Just after I generated the pr there were a few code changes and refactirs applied to the main branch I noticed. I pulled them in and fixed it up and pushed up another commit. Possibly I missed something.

It's still a draft pr -(or it was at least) as I haven't finished doco or tests yet. Was just looking fir initial feedback.

I'll have a look through and retest see what I've missed.

Thanks again

fridaystreet:
@kagux I just reran it again this morning and it all seems to be working for me. Both with max-tokens set and without.

Also tried with a folder specified and without. 

Can you post up the command line you're using or the config.json, not sure what's going on, but sure we can get to the bottom of it.

Cheers

fridaystreet:
I've just noticed something I've missed though. It's not taking into account the token size of the header. I'll have to add some mechanism in to generate the header first and calulate it's size, then iterate over the files with the remaining allowable tokens from maxTokens-headerTokens

fridaystreet:
Also some interesting responses when trying to query the content to see if it's understanding it properly.

Based on the provided context, I cannot determine if the last file element node is the absolute last one in the entire file.

Here's why:

Partial File: The context states that the XML file you shared is only "part 1 of 62". This means there are 61 other files with potentially more file element nodes.

Seems to be getting confused how to interpret the repo size section. I might have to come up with a clearer way of defining the difference between parts and file count in each part.

<repository_size>
This file is part 1 of 62 of a split representation of the entire codebase.
This file contains 33 out of a total of 1054 files.

</repository_size>

PR #128, opened 2024-10-15
feat: Create brew.yml for auto-update of repopack formula

r-dh:
Automatically push new updates to brew.

This will not work without the the COMMITTER_TOKEN secret. 

See https://github.com/mislav/bump-homebrew-formula-action

### DIFF
--- a/.github/workflows/brew.yml
+++ b/.github/workflows/brew.yml
@@ -1,0 +1,17 @@
+name: Brew
+on:
+  push:
+    tags: 'v*'
+
+jobs:
+  brew:
+    name: Bump Homebrew formula version
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


yamadashy:
@r-dh 
Thank you so much for this PR! 
I wasn't familiar with that action. I'll check it out when I get back home. Really appreciate your contribution!

yamadashy:
@r-dh 
Thank you so much!
I've checked the bump-homebrew-formula-action you referenced. This change looks great and doesn't seem to have any issues!
It's fantastic that we won't need to rely on Homebrew maintainers for manual updates anymore. This is definitely a positive improvement! 

Thank you for sharing your knowledge and for this PR!

PR #143, opened 2024-10-30
fix(processConcurrency): improve process concurrency count

twlite:
This method returns an estimate of the default amount of parallelism a program should use, compared to `cpus().length` which returns the cpu core count.


yamadashy:
Hi, @twlite
Thanks for the contribution!

I wasn't aware of os.availableParallelism() - this is an excellent improvement that enables more accurate process concurrency control.

The implementation looks good, so I'll go ahead and merge this PR.

PR #144, opened 2024-10-30
fix: use async setTimeout for sleep

twlite:
```ts
const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
```

This type of hack is not needed since node comes with promise based timers module which can be used as:

```ts
import { setTimeout } from 'node:timers/promises'

await setTimeout(1000)
```

### DIFF
--- a/src/core/packager.ts
+++ b/src/core/packager.ts
@@ -1,11 +1,11 @@
 import fs from 'node:fs/promises';
 import path from 'node:path';
+import { setTimeout } from 'node:timers/promises';
 import pMap from 'p-map';
 import pc from 'picocolors';
 import type { RepopackConfigMerged } from '../config/configTypes.js';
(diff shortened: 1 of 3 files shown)


yamadashy:
@twlite 
Thank you for this great PR!

I wasn't aware of the Promise-based setTimeout in the node:timers/promises module. This is a much cleaner approach than our custom implementation.

The changes look good and everything is working well. Could you just run npm run lint one more time to ensure everything is formatted correctly?

After that, I'll merge this in. Thanks again for the improvement!

twlite:
done, thanks @yamadashy

yamadashy (approved):

yamadashy:
Thank you for the contribution!
Feel free to open more PRs if you find other areas for improvement.

PR #227, opened 2024-12-30
ci(docker): Add release permissions to docker workflow

yamadashy:
ref
- https://github.com/yamadashy/repomix/pull/222#issuecomment-2565003641

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,6 +18,10 @@ on:
   release:
     types: [published, edited]
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   build-and-publish-image:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -17,6 +17,9 @@ on:
   workflow_dispatch:
   release:
     types: [published, edited]
+    permissions:
```
gaby: It's a top-level key, it doesnt go inside this section

yamadashy: My apologies, you're absolutely right! I've corrected it now with a force push.

Issue #282, opened 2025-01-10
XML Escaping


atollk:
I like the feature of having a structured output via XML. However, the string concatenation as it is done in `xmlStyle.ts` doesn't really work. XML requires escaping of certain characters: https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

Basically, if the repo you are reading with repomix contains an XML file itself, it will be parsed as part of the repo tree, not as file contents. The "&&" symbol that is used for boolean conjunction in many languages is disallowed in XML files entirely and will make the output invalid.

The proper solution would be to use an XML serializer that takes care of any escaping that is needed.

atollk:
I'd be fine with submitting a PR for this but it probably requires some refactoring of src/core/output, so I'd rather check with your first.

yamadashy:
Hi, @atollk !
Thank you for your suggestion regarding XML character escaping!

Regarding XML character escaping, I experimented with it in the past but ultimately chose not to implement it for a few reasons:

1. Escaping would increase token count, which is a concern when working with LLM context limits
2. In my use cases (mainly ts, js, md, vue files), LLMs have no issues parsing the unescaped content correctly

However, I agree that proper XML escaping might be useful for some users. We could consider adding this as an optional feature.

atollk:
Hey. Makes sense to me. Just FYI, I encountered this issue when trying to implement RAG.

If it's behind a new flag anyway, I'd propose to alter the output slightly as well to make it parseable as a single XML document. As it currently stands, you always have to manually remove the first line ("This file is a merged...") and add surrounding tags around the rest.

yamadashy:
@atollk 
Thank you for creating this PR!

I hadn't tried RAG myself, so this was a blind spot for me.
The direction looks really good - I particularly like the `parsableStyle` naming, as it's flexible enough to support other output styles like markdown in the future, not just XML.

Would you mind also adding documentation about the new `parsableStyle` option to the README?

yamadashy:
@atollk 
I've just released v0.2.21 with your XML escaping improvements:
https://github.com/yamadashy/repomix/releases/tag/v0.2.21

I'll handle the website updates tomorrow. Thank you for your contribution!

yamadashy:
@atollk 
Oh, I'm sorry for forgetting to close this! 🙏

Thank you again for your contribution to improve XML escaping.
Looking forward to working with you on future improvements!

PR #287, opened 2025-01-12
XML Escaping

atollk:
See #282 

Adds a `parsableStyle` flag to ensure the output strictly follows the specification of the output style.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,8 @@
 [![codecov](https://codecov.io/github/yamadashy/repomix/graph/badge.svg)](https://codecov.io/github/yamadashy/repomix)
 
 📦 Repomix is a powerful tool that packs your entire repository into a single, AI-friendly file.  
-It is perfect for when you need to feed your codebase to Large Language Models (LLMs) or other AI tools like Claude, ChatGPT, and Gemini.
+It is perfect for when you need to feed your codebase to Large Language Models (LLMs) or other AI tools like Claude,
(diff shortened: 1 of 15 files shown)


atollk:
Not yet done: Adding the new option to the website; making the Markdown output parsable as well, for consistency.

atollk:
@yamadashy I'll also add a checkbox with the new feature to the website, unless there is a reason not to?

yamadashy:
@atollk 
> @yamadashy I'll also add a checkbox with the new feature to the website, unless there is a reason not to?
Yes, you can add the checkbox to the website, but there are a few steps we need to follow:

1. First, we need to release this feature to npm, as repomix.com uses the npm package version
2. After the release, I'll need to update the website myself as I'm the only one who can deploy changes

So let's focus on merging and releasing this feature first, and then I'll handle the website update. Sorry for the complicated process!

atollk:
Makes sense to me. I added the backend logic for the website already, which shouldn't be an issue, because it's basically invisible to users. Unless you want me to revert that, I think this is ready to be merged.

yamadashy:
@atollk 
Thank you for the implementation!
I made a few minor tweaks, but the implementation is solid and I've tested it locally with no issues.
Thanks for the refactoring as well!

I'll be releasing a new version soon, followed by server deployment and UI updates. Thank you!

PR #302, opened 2025-01-20
Handle large files gracefully

slavashvets:
Fixes critical out-of-memory errors when processing repositories containing large files

### The Problem
When accidentally adding a large file to a repository (such as a 480MB txt file), Repomix would try to load it into memory and crash with a Node.js heap out-of-memory error:

```
❯ repomix

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
[1]    63614 abort      repomix
```

This provides a poor user experience as:
1. The error is not user-friendly or actionable
2. The program crashes instead of gracefully handling the situation
3. Users don't know which file caused the problem

### The Solution
Added a size limit check (50MB) before loading files into memory, with clear user guidance:

```
📦 Repomix v0.2.21

⚠️ Large File Warning:
────────────────────────────────────────────────
File exceeds size limit: 486.8MB > 50MB (/path/to/large/file)
Add this file to .repomixignore if you want to exclude it permanently

✔ Packing completed successfully!
```

### Changes
- Added 50MB file size limit to prevent memory issues
- Added user-friendly warning message that:
  - Shows the file size and limit
  - Points to the specific file causing the issue
  - Suggests adding it to .repomixignore

### Why
Large files are often added by accident (logfiles, lock files, etc.). Instead of crashing, Repomix should help users identify and handle these files appropriately. The 50MB limit was chosen because:
1. It's well above typical source code file sizes
2. It matches GitHub's recommended file size limit
3. It ensures stable memory usage even on systems with limited RAM

### DIFF
--- a/src/core/file/fileCollect.ts
+++ b/src/core/file/fileCollect.ts
@@ -34,7 +34,7 @@ export const collectFiles = async (filePaths: string[], rootDir: string): Promis
 const readRawFile = async (filePath: string): Promise<string | null> => {
   try {
     const stats = await fs.stat(filePath);
-    
+
     if (stats.size > MAX_FILE_SIZE) {
       const sizeMB = (stats.size / 1024 / 1024).toFixed(1);
       logger.log('⚠️ Large File Warning:');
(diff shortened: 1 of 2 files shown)


yamadashy:
Hi, @slavashvets !
Thank you again for the PR! Really appreciate your continued contributions.

You're right that we don't need to process such large files.
Code looks perfect with minor adjustments I've made.

Going to merge this. Looking forward to your future contributions!

PR #340, opened 2025-02-07
readme: Mention the f2c alternative written in Go

cristim:
I just heard about repomix, a while ago I built a very similar tool in Go for my own use cases (funnily started working on it 10 days before the repomix first commit), mainly focusing on the clipboard operation, which is particularly useful for interactive use.

I recently also updated it to make it aware of Go projects using the Go AST package, copying recursively all the function and type definition dependencies of a given Go function to the clipboard.

## Checklist

No, since this is just a readme change.

- [ ] Run `npm run test`
- [ ] Run `npm run lint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ If you're using Python, you might want to check out `Gitingest`, which is better
 science workflows:
 https://github.com/cyclotruc/gitingest
 
+The [f2c](https://github.com/LeanerCloud/f2c) tool is a more lightweight alternative written in Go that only supports the clipboard, without generating files. For Go projects it supports recursively looking into the current Go package using the Go AST to copy only a function's dependencies to the clipboard.
+
 ## 📊 Usage
 
 To pack your entire repository:


yamadashy:
Hi @cristim,

Thank you for sharing about f2c! I appreciate you taking the time to contribute this information.

I want to clarify why we currently list only [Gitingest](https://github.com/cyclotruc/gitingest) in our Alternative Tools section. We've received many requests from Python users looking for a Python-based solution, while Gitingest users have been asking for a JavaScript/TypeScript alternative. This mutual need led us to establish a collaboration where we can each address these user requests by recommending each other's tools.

While f2c looks like an interesting tool with its own unique features around Go AST handling, I currently want to maintain this focused approach based on our users' specific needs. That said, I'm always open to discussing potential collaborations if you see similar complementary opportunities between our user bases!

Thanks again for the suggestion and for sharing your work!
It's great to see others building interesting tools in this space.

yamadashy:
Based on my previous explanation about our current focused approach, I'll be closing this PR.
Thank you again for sharing f2c with us!

PR #414, opened 2025-03-12
Update README.md

eapo:
1. Separate installation scripts
2. Add NixOS shell.nix method

---

#### Alternatively using shell.nix (NixOS)
```bash
{ pkgs ? import <nixpkgs> { } }:

pkgs.mkShell {
  buildInputs = [
  ];
 
  shellHook =
    ''
      npx repomix --ignore "data/,ai-docs/,myenv/,.vscode/"
    '';
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,17 +62,38 @@ npx repomix
 
 Or install globally for repeated use:
 
+#### Install using npm
 ```bash
-# Install using npm
 npm install -g repomix
+```
 
-# Alternatively using yarn
+#### Alternatively using yarn
+```bash
 yarn global add repomix
+```
 
-# Alternatively using Homebrew (macOS/Linux)
+#### Alternatively using Homebrew (macOS/Linux)
+```bash
 brew install repomix
+```
 
-# Then run in any project directory
+#### Alternatively using `shell.nix` (NixOS)
+```bash
+{ pkgs ? import <nixpkgs> { } }:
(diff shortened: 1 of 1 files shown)


yamadashy:
Hi, @eapo !
Thank you for your PR! I appreciate you adding installation instructions for NixOS users.

I think this is a valuable addition, but I'd like to suggest a small change to the placement. Since the `shell.nix` approach isn't strictly a "global installation" but rather a project-specific environment setup, it might be confusing to include it in the "Install globally for repeated use" section.

Would it be possible to create a new section titled "Using with NixOS" and place it before the "Alternative Tools 🛠️" section? This would make the README structure clearer and help users find the installation method appropriate for their environment.

A structure like this would work well:

```markdown
## Using the CLI Tool `>_`

### Install globally for repeated use:
...

### Using with NixOS
For NixOS users, you can create a `shell.nix` file in your project:
...

## Alternative Tools 🛠️
...
```

What do you think about this organization? It clearly separates the global installation section from the NixOS usage method.

Thanks again for contributing to make Repomix more accessible to NixOS users!

yamadashy:
Hi, @eapo !
Quick update — repomix is now in nixpkgs, so the install path is much simpler now (`nix profile add nixpkgs#repomix`). I'll close this and add a Nix section to the README separately 🙏

Issue #438, opened 2025-03-27
Feature Request: Add a "force include" configuration/CLI option to override default and custom exclusion rules


rockmandash:
I frequently need to quickly bundle just a few specific files from my repository using `repomix.config.json`—particularly when I'm repeatedly adjusting the `include` field. However, I'm running into an issue: I'd like to include a file located in `node_modules/`, but Repomix automatically excludes that folder by default. Even if I specify the exact path to the file inside `node_modules/`, it’s not included because of the default or custom exclusion rules.

It would be incredibly helpful to have a **"force include"** option (either in the configuration file or as a CLI flag) that explicitly overrides all exclusion rules—both the default `node_modules` exclusion and any patterns set by `.gitignore`, `.repomixignore`, or other custom rules. This way, if I really need to pull in a specific file from `node_modules/` (or any other traditionally excluded directory), I can do so without having to disable or modify all the exclude patterns.

---

**Use Case**  
- Quickly bundling only a few files for AI analysis without including the entire node_modules directory.  
- Ensuring that even if `node_modules` or similar directories are globally excluded, certain critical files can still be selectively included.  
- Reducing back-and-forth toggling of ignore patterns in `repomix.config.json`.

---

**Proposed Solution**  
1. **Configuration File Approach**:  
   - Add a new `forceInclude` array (similar to `include` and `ignore`) in `repomix.config.json` that overrides all exclusions.  
   - Example:
     ```json
     {
       "forceInclude": [
         "node_modules/some-package/dist/specific-file.js"
       ]
       // ... other config properties
     }
     ```

2. **CLI Flag Approach**:  
   - Provide a `--force-include` option that can be used once or multiple times to specify paths.  
   - Example:
     ```bash
     repomix --force-include "node_modules/some-package/dist/specific-file.js"
     ```

3. **Behavior**:  
   - Any pattern or path listed under `forceInclude` would be packed regardless of any exclusion rules set by `.gitignore`, `.repomixignore`, default ignore patterns, or the `ignore.customPatterns` configuration.

---

**Benefits**  
- Simplifies the process when a single or handful of files from an otherwise excluded directory need to be included.  
- Makes the configuration more flexible by letting users override strict defaults without disabling them entirely.  
- Reduces repetitive manual toggling of ignore patterns.

---

**Additional Notes / Considerations**  
- This feature might also be useful for teams that store partial dependencies or patched versions inside `node_modules` or similarly excluded folders.  
[...]

PR #480, opened 2025-04-12
Feature: Add --force-include option

ktyubeshi:
## Description:
This PR adds a `--force-include` feature to include specific files regardless of ignore patterns.

### Key changes:
- Added functionality to force include files by overriding ignore patterns
- Implemented command-line option `--force-include <patterns>`
- Added `forceInclude` property to configuration files
- Updated related tests and documentation

### Usage example:
```bash
repomix --force-include "coverage/summary.json,test-results/junit.xml"
```

This feature allows including normally ignored files (e.g., coverage reports or specific log files) in the output when needed.

### Documentation
- Updated README.md with the new option
- Updated website documentation in both English and Japanese versions
- Added explanation of ignore pattern priority in configuration guides

### Testing
- Added tests for the new feature
- All existing tests pass with the new implementation

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`


yamadashy:
Hi, @ktyubeshi !
Thanks so much for the PR!

I assume this is intended to address a similar use case as mentioned in this issue:
https://github.com/yamadashy/repomix/issues/438

I totally understand the motivation here — and I really appreciate the effort you've put into the implementation.
That said, as I mentioned in the [issue comment](https://github.com/yamadashy/repomix/issues/438#issuecomment-2816687383), I'm still a bit hesitant about introducing a `--force-include` option at this point. My current thinking is to support these scenarios through a `useDefaultPatterns: false` option instead.

Also, the use of the word "force" might create long-term challenges from a design perspective. For example, if users want to "force include" a set of files but then also ignore some of them, it becomes difficult to reason about what should happen.

I’d love to explore a more flexible and conceptually cleaner solution before locking in this approach. Thanks again for opening the PR — it’s a helpful contribution to the discussion!

ktyubeshi:
Hi, @yamadashy 

Thank you for your feedback.
I wasn't aware of #438  when implementing this, but I think our use cases and ideas are essentially identical.

> Also, the use of the word "force" might create long-term challenges from a design perspective.

Regarding the choice of the word "force," my intention was to use a strong expression to clearly indicate that this would be the highest priority.
When adding this feature, I was also concerned that carelessly adding options might cause confusion, and I initially tried to find a solution using only the existing ignore settings. However, I added `force include` because the configuration would become too complicated, and this could be achieved with a simple setting.
I understand the lingering concerns about whether this is truly the best choice. I agree that careful consideration is needed.

> For example, if users want to "force include" a set of files but then also ignore some of them, it becomes difficult to reason about what should happen.

Regarding this, while an option like --force-include-ignore could be considered, I thought it might be acceptable to simply decide against adding such functionality.

In my actual use cases, the targets I would specify with `--force-include` are things like:
* Test results
* Error logs
* Static analysis tool results
These files are mostly either single files or exist together in specific folders, so I don't think there's a need for complex specifications that would filter them further with ignore rules.

> My current thinking is to support these scenarios through a useDefaultPatterns: false option instead.

The reason I added this option was that I wanted to keep repomix's default excluded files as they are.
If we set `useDefaultPatterns` to false as you suggested, I think we would need to manually reconfigure everything.
I'd be happy if there's a good way to achieve the same thing with minimal configuration.

PR #529, opened 2025-05-05
Put link on the text in github-actions.md

szepeviktor:
No more _Click here_.


yamadashy:
@szepeviktor 
Thanks always!
I’ve also updated the other languages.
Merging this now 🙌

PR #533, opened 2025-05-06
feat(output): Add git diff support with --diffs flag

pmdyy:
Added git diff support with --diffs flag; showing working tree diffs. 

I think this is helpful when you are iterating against a feature/issue, so that the LLM is aware of the changes you made so far, until you've staged it. Not really sure where to put the `diffs` section, but for now I've placed it before instruction (if available).

## Checklist

- [ X] Run `npm run test`
- [ X] Run `npm run lint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,7 @@ Instruction
 - `--header-text <text>`: Custom text to include in the file header
 - `--instruction-file-path <path>`: Path to a file containing detailed custom instructions
 - `--include-empty-directories`: Include empty directories in the output
+- `--include-diffs`: Include git diffs in the output (includes both work tree and staged changes separately)
 - `--no-git-sort-by-changes`: Disable sorting files by git change count (enabled by default)
 
 #### Filter Options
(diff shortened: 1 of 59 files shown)


yamadashy:
@pmdyy 
Thank you so much for your contribution!

I appreciate the thorough implementation including all the tests.

I've made a few adjustments:
* Added staged changes to the git diff output in addition to working tree changes. This makes it more useful for considering commit messages and evaluating commit granularity.
* Renamed the option to `--include-diffs` instead of just `--diffs` to avoid confusion with other potential features.
* Refactoring:
  * Storing processing information in the config isn't ideal, so I've created a `GitDiffResult` object to pass around instead.

I'll review this a bit more before merging.

yamadashy:
/gemini review
@coderabbitai review

yamadashy:
@pmdyy
I’ve since pushed a few more tweaks and fixes on top of your original changes:

* Ran secretlint against the git diffs as an extra precaution
* Updated the documentation to cover the new `--include-diffs` flag

Everything is looking good now and all tests are passing, so I’ll go ahead and merge this.

Thanks again for your awesome work!

PR #556, opened 2025-05-15
feat: --no-file-summary now also suppresses generationHeader; doesn't suppress headerText (fixes #554)

### LINKED ISSUE #554: Feature Request: `--no-file-summary` should also suppress `generationHeader`; should not suppress user-specified `headerText`
Repomix's output has a [summary/usage guidelines section](https://github.com/yamadashy/repomix/blob/fbad7c9ae364e52eb0cd153c451fc54f1606850a/src/core/output/outputGenerate.ts#L61), which can be suppressed with `--no-file-summary`,  but it also has a [generationHeader](https://github.com/yamadashy/repomix/blob/fbad7c9ae364e52eb0cd153c451fc54f1606850a/src/core/output/outputGenerate.ts#L60) section which cannot be suppressed. 

My use case is for packing files other than source code (project documentation, code guidelines, etc), but the text in the unsuppressable [generationHeader](https://github.com/yamadashy/repomix/blob/fbad7c9ae364e52eb0cd153c451fc54f1606850a/src/core/output/outputStyleDecorate.ts#L50) has static references to "the codebase", which may mislead the reader of the file (especially since I'm lazy and just pass the directory name in argv instead of using `--include`, so the output says it's "the entire codebase"). 

Additionally, the `--header-text` config item adds its contents to the [summary/usage guidelines section](https://github.com/yamadashy/repomix/blob/7d1aa52697e2fe4342a21a4607254e5d84e043f8/src/core/output/outputStyleDecorate.ts#L118). When that section is suppressed with `--no-file-summary`, the custom `--header-text` is suppressed along with it. 

Desired behavior:

- `--no-file-summary` should suppress both the summary/usage guidelines section and the `generationHeader`
- `--header-text`, if provided, should be included in the output, regardless of the `--no-file-summary` flag

joshwand:
--no-file-summary should also suppress generationHeader; should not suppress user-specified headerText ( fixes #554 )

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/core/output/outputGenerate.ts
+++ b/src/core/output/outputGenerate.ts
@@ -62,7 +62,7 @@ const generateParsableXmlOutput = async (renderContext: RenderContext): Promise<
             '#text': renderContext.generationHeader,
             purpose: renderContext.summaryPurpose,
             file_format: `${renderContext.summaryFileFormat}
-4. Repository files, each consisting of:
+5. Repository files, each consisting of:
   - File path as an attribute
   - Full contents of the file`,
             usage_guidelines: renderContext.summaryUsageGuidelines,
--- a/src/core/output/outputGenerate.ts
+++ b/src/core/output/outputGenerate.ts
@@ -57,10 +57,9 @@ const generateParsableXmlOutput = async (renderContext: RenderContext): Promise<
   const xmlBuilder = new XMLBuilder({ ignoreAttributes: false });
   const xmlDocument = {
     repomix: {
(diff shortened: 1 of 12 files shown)


yamadashy:
@joshwand 
Thank you for addressing this!

I made a few minor tweaks, and everything looks good!
I’ll merge this now.

Please let me know if there’s anything else.

yamadashy (approved):

PR #682, opened 2025-06-22
feat(output): Make summary purpose conditional based on file selection

eriknygren:
## What
- Update generateSummaryPurpose function to accept RepomixConfigMerged parameter
- Add logic to differentiate between entire codebase vs subset in purpose description
- Update function call in outputGenerate.ts to pass config parameter
- Add comprehensive tests for both entire codebase and subset scenarios

## Why
Previously, the summary purpose section always claimed the file contained "the entire repository's contents", even when repomix was configured to process only a subset of files (e.g., with include patterns). This created inconsistency with other parts of the output that correctly indicated when only a subset was processed.

This change ensures the purpose section accurately describes the content as either "the entire repository's contents" or "a subset of the repository's contents that is considered the most important context" based on the actual file selection configuration.

Let me know what you think! I think this removes a potential "contradiction" for the LLMs parsing this file, and thanks for an awesome repo.

## Checklist

- [ x] Run `npm run test`
- [ x] Run `npm run lint`

### DIFF
--- a/src/core/output/outputGenerate.ts
+++ b/src/core/output/outputGenerate.ts
@@ -31,7 +31,7 @@ const calculateMarkdownDelimiter = (files: ReadonlyArray<ProcessedFile>): string
 const createRenderContext = (outputGeneratorContext: OutputGeneratorContext): RenderContext => {
   return {
     generationHeader: generateHeader(outputGeneratorContext.config, outputGeneratorContext.generationDate),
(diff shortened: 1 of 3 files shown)


yamadashy:
Hi, @eriknygren !
Thank you for the PR!

I hadn't noticed this issue.
Looks good to me, so I'll merge it. Thanks again for your contribution!

PR #727, opened 2025-07-11
feat: Add originalLineNumbers option for line number placement control

dorianjp:
# Add originalLineNumbers option for line number placement control

## Overview

This PR introduces a new configuration option `originalLineNumbers` and CLI flag `--output-original-line-numbers` that allows users to control when line numbers are added to the content during processing. When enabled, line numbers are added to the original content before any transformations (like comment removal, empty line removal, or compression) are applied.

## Changes

1. **Configuration Schema**
   - Added `originalLineNumbers` to the configuration schema in `src/config/configSchema.ts`
   - Set default value to `false`

2. **CLI Option**
   - Added `--output-original-line-numbers` CLI option in `src/cli/cliRun.ts`
   - Updated the `CliOptions` interface in `src/cli/types.ts`
   - Added handling for the option in `src/cli/actions/defaultAction.ts`

3. **Processing Logic**
   - Refactored the file processing logic in `src/core/file/fileProcessContent.ts`:
     - Extracted line number addition logic into a reusable `addLineNumbers` function
     - Added conditional logic to add line numbers at different stages based on the `originalLineNumbers` setting
     - Changed the order of operations to ensure line numbers are added before or after transformations as appropriate

4. **Documentation**
   - Updated the configuration documentation in `website/client/src/en/guide/configuration.md`
   - Updated the command line options documentation in `website/client/src/en/guide/command-line-options.md`
   - Updated the README.md with the new CLI option

5. **Tests**
   - Added tests for the `originalLineNumbers` feature in `tests/core/file/fileProcessContent.test.ts`
   - Added tests for the CLI option in `tests/cli/actions/defaultAction.test.ts`
   - Updated the schema tests in `tests/config/configSchema.test.ts`
   - Fixed the failing tests in `tests/core/file/fileProcess.test.ts`

## Feature Behavior

The `originalLineNumbers` feature works as follows:

1. **When `showLineNumbers` is false**:
   - No line numbers are added, regardless of the `originalLineNumbers` setting

2. **When `showLineNumbers` is true and `originalLineNumbers` is false**:
   - Line numbers are added after all transformations (compression, comment removal, empty line removal)
   - This is the previous default behavior

3. **When `showLineNumbers` is true and `originalLineNumbers` is true**:
   - Line numbers are added to the original content before any transformations
   - Transformations are then applied to the content with line numbers

## Benefits

1. **Preserves Original Line Numbers**: When using features like comment removal or empty line removal, the original line numbers are preserved, making it easier to reference specific lines in the original file.

2. **Better Compression Integration**: When using compression with line numbers, the line numbers are included in the compressed output, providing context for the compressed code.

3. **Improved Readability**: Line numbers added before transformations can make it easier to understand how the transformations affected the content.

### Configuration File
```json
{
  "output": {
    "showLineNumbers": true,
    "originalLineNumbers": true,
    "removeComments": true
  }
}
```

### CLI Command
```bash
repomix --output-show-line-numbers --output-original-line-numbers --remove-comments
```

## Testing

All tests are passing. The implementation has been tested with various combinations of options to ensure it works correctly in all scenarios.

### DIFF
--- a/src/core/file/fileProcessContent.ts
+++ b/src/core/file/fileProcessContent.ts
@@ -23,9 +23,7 @@ export const processContent = async (rawFile: RawFile, config: RepomixConfigMerg
 
   logger.trace(`Processing file: ${rawFile.path}`);
 
-  const originalLineNumbers = config.output.originalLineNumbers ?? false;
-
-  if (config.output.showLineNumbers && originalLineNumbers) {
+  if (config.output.showLineNumbers && config.output.originalLineNumbers) {
     processedContent = addLineNumbers(processedContent);
   }
 
@@ -53,7 +51,7 @@ export const processContent = async (rawFile: RawFile, config: RepomixConfigMerg
     }
   }
 
-  if (config.output.showLineNumbers && !originalLineNumbers) {
+  if (config.output.showLineNumbers && !config.output.originalLineNumbers) {
     processedContent = addLineNumbers(processedContent);
   }
 


yamadashy:
@dorianjp 
Thank you for the PR!
Is this the implementation for what DanielusG and I were discussing?

I've made some minor adjustments to the code.
There are still a few things I'd like to adjust, so I'll work on them tomorrow or later.

Regarding the option name, since this option is dependent on `showLineNumbers`, I'm thinking of renaming it to `showLineNumbersPreserveOriginal` even though it's a bit long.

yamadashy:
I need to be cautious here since I'm not sure if compression (tree-sitter) will work correctly with this implementation.

My initial assumption was that we would need to properly insert the original line numbers after compression within the compress function itself.

dorianjp:
> I need to be cautious here since I'm not sure if compression (tree-sitter) will work correctly with this implementation.
> 
> My initial assumption was that we would need to properly insert the original line numbers after compression within the compress function itself.

Good call, I never tested that since I don't use the compression aspect. I found it speeds up LLM comprehension and edits by a huge factor with original line numbers, in fact some of the 'dumb' LLMs are effective now with accurate line numbers to go on, whereas before without line numbers or misleading line numbers they would just YOLO random shit or break similar but not the same code blocks for intentional edits. The 'new' line numbering actually trips up a few LLMs, which is why I made this PR.

PR #947, opened 2025-11-11
fix(config): set file extension based on style parameter

Ahmad8864:
## Summary

This PR updates the logic in mergeConfigs to handle output file paths more consistently. The output file name now automatically matches the selected style, unless a specific file path is set in the config or via the CLI.

## Changes
* **Config merging**: 
Updated mergeConfigs in src/config/configLoad.ts so that when only a style is provided, the default output file path updates to match that style. If a file path is explicitly set, it stays unchanged.
* **Tests**: 
Added new tests in tests/config/configLoad.test.ts to cover these scenarios and verify that the path updates correctly when it should, and stays the same when a file path is explicitly defined.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/config/configLoad.ts
+++ b/src/config/configLoad.ts
@@ -195,6 +195,7 @@ export const mergeConfigs = (
         },
       };
 
+      // Auto-adjust filePath extension to match style when filePath is not explicitly set
       const style = mergedOutput.style ?? baseConfig.output.style;
       const filePathExplicitlySet = Boolean(fileConfig.output?.filePath || cliConfig.output?.filePath);
 


Ahmad8864:
@yamadashy Hi there, just wanted to check if I needed to do anything else to get a review.

yamadashy:
Hi, @Ahmad8864!
Thank you so much for this fix PR!

It looks like there was a regression introduced in #923. I really appreciate you catching this and fixing it!

The code looks good. I just added a comment to explain the logic.

I'll merge this now. Thanks again!

PR #965, opened 2025-11-24
fix (actions) remove no supported --no-compress  argument

chinchala:
This PR fixes a GitHub Actions bug where passing `compress: false` incorrectly added the `--no-compress` argument, which is not a supported flag.
The logic is now corrected so that the `--compress` argument is added only when `compress: true` is provided.


yamadashy:
Hi @chinchala !
Thank you for the PR!

I completely missed this issue during implementation.

The changes look good, so I'll merge this! Feel free to contribute again!

PR #1078, opened 2026-01-06
fix: add timeout and disable terminal prompts for remote git commands

### LINKED ISSUE #1077: pack_remote_repository hangs indefinitely on inaccessible repos
### Description

When using `pack_remote_repository` (MCP) or `--remote` (CLI) with an inaccessible repository (non-existent, private, or requiring auth), repomix hangs indefinitely.

### Root Cause

`execLsRemote` in `src/core/git/gitCommand.ts` has no timeout and doesn't set `GIT_TERMINAL_PROMPT=0`. Git waits for credential input that never comes in non-interactive contexts.

### Reproduce

```bash
npx repomix --remote https://github.com/user/nonexistent-repo
```

### Expected

Fail with error after reasonable timeout.

### Fix

```typescript
const result = await deps.execFileAsync('git', ['ls-remote', ...], {
  timeout: 30000,
  env: { ...process.env, GIT_TERMINAL_PROMPT: '0' }
});
```

Apply to all git commands in gitCommand.ts.

Pipboyguy:
Prevents indefinite hangs when accessing inaccessible repositories (non-existent, private, or requiring auth).

Changes:
- Set `GIT_TERMINAL_PROMPT=0` to fail fast instead of waiting for credentials
- Add 30s timeout to remote git operations

Fixes #1077

### DIFF
--- a/src/core/file/fileTreeGenerate.ts
+++ b/src/core/file/fileTreeGenerate.ts
@@ -113,3 +113,80 @@ export const generateTreeStringWithLineCounts = (
   const tree = generateFileTree(files, emptyDirPaths);
   return treeToStringWithLineCounts(tree, lineCounts).trim();
 };
+
+/**
+ * Represents files grouped by their root directory.
+ */
+export interface FilesByRoot {
+  rootLabel: string;
+  files: string[];
+}
+
+/**
+ * Internal helper function to generate multi-root tree sections.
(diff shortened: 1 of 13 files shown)


yamadashy:
Hi @Pipboyguy!

Thanks for the PR! The fix looks good and the code is solid.

Merging this now. Thanks for the contribution! 🙏

PR #1160, opened 2026-02-20
fix(cli): Respect local config when using --remote option

### LINKED ISSUE #1025: Repomix local config breaks when using `--remote` option
## Bug / Unintuitive DX
Running `repomix --remote https://github.com/user/repo.git` ignores the local repomix.config.json in my workspace because repomix first clones the remote repo and then searches for the config in the target repo instead of my current working directory.

Specifying a custom config with `-c myconfig.json` also initially seems to fail as well.

## Solution
Technically, you can get the custom config working with an explicity path as opposed to a relative one but this is a bit unintuitive.
`repomix --remote https://github.com/user/repo.git -c "$PWD/repomix.config.json"`

The challenge is there are 3 potentiall options to pull config from
1. Global config (home directory)
2. Local workspace (local directory)
3. Target repo (remote repo)

Since the purpose of the config file is for the library consumer rather than the library author it would be better if repomix used the local config instead of search the target repo for a config file.

### Usage Context

Repomix CLI

### Repomix Version

1.11.0

### Node.js Version

v22.19.0

hztBUAA:
Fixes #1025

When using `--remote`, repomix ignores the local `repomix.config.json` because it searches for config files in the cloned temporary directory instead of the current working directory (CWD).

This PR resolves the issue by ensuring config paths are resolved from CWD before being passed to `runDefaultAction`:

1. **Explicit `-c` config paths** are resolved to absolute paths relative to CWD, so they remain valid even when the working directory changes to the temp clone directory.
2. **Auto-discovery** searches CWD for default config files (e.g., `repomix.config.json`) first via the new `findLocalConfigPath` utility, and passes the found absolute path to `runDefaultAction`.
3. If no local config is found in CWD, the behavior falls through to the existing global config search logic.

### Changed files

- `src/cli/actions/remoteAction.ts` — Resolve config from CWD before calling `runDefaultAction`
- `src/config/configLoad.ts` — Export new `findLocalConfigPath` helper
- `src/index.ts` — Add `findLocalConfigPath` to public API exports
- `tests/cli/actions/remoteAction.test.ts` — Add 3 tests for the config resolution behavior

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1160" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/cli/actions/remoteAction.ts
+++ b/src/cli/actions/remoteAction.ts
@@ -2,6 +2,7 @@ import * as fs from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
 import pc from 'picocolors';
+import { findLocalConfigPath } from '../../config/configLoad.js';
 import { execGitShallowClone } from '../../core/git/gitCommand.js';
 import { downloadGitHubArchive, isArchiveDownloadSupported } from '../../core/git/gitHubArchive.js';
 import { getRemoteRefs } from '../../core/git/gitRemoteHandle.js';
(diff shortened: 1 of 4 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

PR #1166, opened 2026-02-20
fix(ignore): use case-sensitive matching to prevent BUILD file exclusion on macOS

### LINKED ISSUE #980: fix(ignore): BUILD files incorrectly ignored on case-insensitive file systems (macOS)
## Summary

On macOS (case-insensitive file system), pants build tool's `BUILD` files are incorrectly ignored because the default ignore pattern `build/**` matches them.

## Problem

- Repomix has `build/**` in [defaultIgnore.ts:111](https://github.com/yamadashy/repomix/blob/main/src/config/defaultIgnore.ts#L111)
- macOS uses a case-insensitive file system by default (HFS+/APFS)
- The `BUILD` files (used by [pants build tool](https://www.pantsbuild.org/)) are matched by `build/**` pattern
- Users cannot simply add `!BUILD` to include patterns because that would still be affected by the ignore

## Expected Behavior

`BUILD` files should be included in repomix output, as they are source configuration files (not build output directories).

## Current Workaround

Users need to disable default patterns entirely and manually specify all other patterns:

```json
{
  "ignore": {
    "useDefaultPatterns": false,
    "customPatterns": [
      // manually add all patterns except build/**
    ]
  }
}
```

This is not ideal.

## Potential Solutions

1. **Explicitly set `caseSensitiveMatch: true`** in globby options at `createBaseGlobbyOptions()` in [fileSearch.ts](https://github.com/yamadashy/repomix/blob/main/src/core/file/fileSearch.ts#L306-L319)

2. **Change the pattern** from `build/**` to something more specific that won't match `BUILD` files

3. **Add `!BUILD` and `!**/BUILD`** to a "never ignore" list that takes precedence

## Environment

- OS: macOS (case-insensitive file system)
- Build tool: pants (uses `BUILD` files)

## Reference

- Discord report: https://discord.com/channels/1324644561959522314/1324644562429542425/1439783115533844521

lucamorettibuilds:
## Summary

On case-insensitive file systems (macOS HFS+/APFS), the default ignore pattern `build/**` incorrectly matches Pants build tool's `BUILD` files. This one-line fix ensures ignore patterns match with exact case.

## Problem

- Default ignore has `build/**` ([defaultIgnore.ts:111](https://github.com/yamadashy/repomix/blob/main/src/config/defaultIgnore.ts#L111))
- macOS HFS+/APFS is case-insensitive by default
- `BUILD` files (used by [Pants](https://www.pantsbuild.org/)) get matched by `build/**`
- Users can't simply add `!BUILD` to include patterns

## Fix

Sets `caseSensitiveMatch: true` in `createBaseGlobbyOptions()`. This is a globby/fast-glob option that forces case-sensitive pattern matching regardless of the OS filesystem. `build/**` will only match lowercase `build/` directories — not `BUILD` source files.

This is the simplest and most correct approach from the potential solutions listed in #980 (option 1). It also matches the behavior users expect from their `.gitignore` and `.repomixignore` patterns.

## Changes

- `src/core/file/fileSearch.ts`: Add `caseSensitiveMatch: true` to `createBaseGlobbyOptions()`
- `tests/core/file/fileSearch.test.ts`: Update globby options assertions

Fixes #980

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1166" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/core/file/fileSearch.ts
+++ b/src/core/file/fileSearch.ts
@@ -316,6 +316,9 @@ const createBaseGlobbyOptions = (
   absolute: false,
   dot: true,
   followSymbolicLinks: false,
+  // Force case-sensitive pattern matching regardless of OS filesystem case sensitivity.
+  // Without this, on macOS (HFS+/APFS case-insensitive), the default ignore pattern
+  // `build/**` would incorrectly exclude Pants BUILD files.
   caseSensitiveMatch: true,
 });
 
--- a/src/core/file/fileSearch.ts
+++ b/src/core/file/fileSearch.ts
@@ -316,6 +316,7 @@ const createBaseGlobbyOptions = (
   absolute: false,
   dot: true,
   followSymbolicLinks: false,
+  caseSensitiveMatch: true,
 });
 
 export const getIgnoreFilePatterns = async (config: RepomixConfigMerged): Promise<string[]> => {
--- a/tests/core/file/fileSearch.test.ts
+++ b/tests/core/file/fileSearch.test.ts
@@ -290,6 +290,7 @@ node_modules
           absolute: false,
           dot: true,
           followSymbolicLinks: false,
+          caseSensitiveMatch: true,
         }),
       );
     });
@@ -943,6 +944,7 @@ node_modules
           absolute: false,
           dot: true,
           followSymbolicLinks: false,
+          caseSensitiveMatch: true,
         });
 
         // Each call should have either onlyFiles or onlyDirectories, but not both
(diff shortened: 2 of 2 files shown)


yamadashy:
Thanks for the fix for #980! 🙏

I checked this end-to-end, and `caseSensitiveMatch: true` turns out to be a no-op: fast-glob already defaults it to `true`, so the output is byte-identical.

I also couldn't reproduce the original problem on current `main` — the default `build/**` pattern doesn't match a file named `BUILD`, so `BUILD` files are already included.

The layer that can actually exclude a `BUILD` file is the gitignore side (a bare `build` entry), which this patch doesn't change. I've left the details on #980.

So I'll close this one for now, but I really appreciate the investigation and contribution! 🙏

PR #1235, opened 2026-03-17
perf(core): Optimize sortPaths with decorate-sort-undecorate pattern

yamadashy:
## Summary

Optimize three performance-critical functions in the file processing pipeline:

1. **`sortPaths`** (filePathSort.ts): Pre-compute `path.split()` results before sorting, avoiding O(N log N) repeated string split allocations in sort comparisons
2. **`generateFileTree`** (fileTreeGenerate.ts): Use `Map` for O(1) child node lookup instead of `Array.find()` O(N) scan; use `array.push()` + `join()` instead of string concatenation in tree stringification; eliminate redundant recursive sorting
3. **`filterOutUntrustedFiles`** (filterOutUntrustedFiles.ts): Use `Set.has()` for O(1) lookup instead of `Array.some()` O(M) linear scan

#### `sortPaths` — **6-7x faster**

| Files | Before | After | Speedup |
|-------|--------|-------|---------|
| 1,000 | 5.27ms | 0.83ms | **6.4x** |
| 5,000 | 33.96ms | 4.88ms | **7.0x** |
| 10,000 | 65.51ms | 10.83ms | **6.0x** |

#### `generateTreeString` — **1.3x faster at scale**

| Files | Before | After | Speedup |
|-------|--------|-------|---------|
| 1,000 | 1.01ms | 1.05ms | ~1.0x |
| 5,000 | 6.27ms | 5.63ms | **1.1x** |
| 10,000 | 16.42ms | 12.53ms | **1.3x** |

#### Tree construction only (Map vs Array.find):
| Files | Array.find | Map | Speedup |
|-------|-----------|-----|---------|
| 10,000 | 9.77ms | 8.32ms | **1.2x** |
| 20,000 | 33.74ms | 18.23ms | **1.9x** |

#### `filterOutUntrustedFiles` — **up to 30x faster**

| Files / Suspicious | Before | After | Speedup |
|--------------------|--------|-------|---------|
| 1,000 / 50 | 0.12ms | 0.04ms | **2.7x** |
| 5,000 / 200 | 2.78ms | 0.20ms | **13.9x** |
| 10,000 / 500 | 12.99ms | 0.43ms | **30.3x** |

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1235" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/tests/core/file/fileSearch.test.ts
+++ b/tests/core/file/fileSearch.test.ts
@@ -531,11 +531,12 @@ node_modules
         },
       });
 
+      const sep = path.sep;
       const mockFileStructure = [
-        'root/file1.js',
-        'root/another/file3.js',
-        'root/subdir/file2.js',
-        'root/subdir/ignored.js',
+        `root${sep}file1.js`,
+        `root${sep}another${sep}file3.js`,
+        `root${sep}subdir${sep}file2.js`,
+        `root${sep}subdir${sep}ignored.js`,
       ];
 
       vi.mocked(globby).mockResolvedValue(mockFileStructure);
@@ -543,12 +544,12 @@ node_modules
       const result = await searchFiles('/mock/root', mockConfig);
 
       expect(result.filePaths).toEqual([
-        'root/another/file3.js',
-        'root/subdir/file2.js',
-        'root/subdir/ignored.js',
-        'root/file1.js',
+        `root${sep}another${sep}file3.js`,
+        `root${sep}subdir${sep}file2.js`,
+        `root${sep}subdir${sep}ignored.js`,
+        `root${sep}file1.js`,
       ]);
-      expect(result.filePaths).toContain('root/subdir/ignored.js');
+      expect(result.filePaths).toContain(`root${sep}subdir${sep}ignored.js`);
       expect(result.emptyDirPaths).toEqual([]);
     });
 


yamadashy:
## Code Review

 already left a thorough review covering the key points. Adding a few incremental observations below.

### AI Bot Inline Comments
No inline review comments from bots — nothing to evaluate.

### Incremental observations

**fileTreeGenerate.ts —  placement**

The refactoring moves  to the public API boundary (, ), which runs it once per call instead of recursively at every level. One thing to be aware of: if a caller ever reuses the same `TreeNode` for both `treeToString` and `treeToStringWithLineCounts`, the tree would be sorted twice. This is harmless (sorting an already-sorted tree is O(N log N) but correct), and in the current codebase every caller creates a fresh tree — so it's not an issue now. Just worth noting as the API evolves.

**filePathSort.ts — `presplit` memory allocation**

For very large repos (e.g., 100k files), `presplit` holds N arrays of path parts in memory simultaneously alongside the original `filePaths` array. Given the sort itself is O(N log N) and `presplit` is O(N \* depth), the memory overhead is proportional to the input size — a reasonable tradeoff for the 6–7x speedup. No change needed, just confirming the tradeoff is intentional.

**filterOutUntrustedFiles.ts**

Clean and correct. The `Set` construction from `suspiciousFilesResults.map()` is O(M) and `has()` is O(1). For the typical case where suspicious files are a small fraction of total files, this is strictly better.

---

Overall: The optimizations are algorithmically sound, benchmarks are compelling, and the implementation matches the intent. The `nodeMap?.set()` optional chaining noted by `claude[bot]` is the only minor nit. **Good to merge.**

PR #1336, opened 2026-03-28
perf(core): Reduce token counting sample and use sqrt-weighted ratio estimation

yamadashy:
## Summary

Full token counting of every file is expensive. This change samples a subset of files for actual token counting and estimates the rest using a sqrt-weighted character-to-token ratio, achieving nearly identical accuracy at a fraction of the cost.

Benchmark: -0.13s consistent across Ubuntu, macOS, and Windows.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/core/metrics/calculateMetrics.ts
+++ b/src/core/metrics/calculateMetrics.ts
@@ -70,6 +70,44 @@ export const getMetricsTargetPaths = (processedFiles: ProcessedFile[], config: R
     .map((file) => file.path);
 };
 
+/**
+ * Compute sqrt-weighted char:token ratio from file metrics.
+ * Sqrt-weighting corrects for size bias: large files tend to have higher
+ * token/char ratios, so weighting by sqrt(charCount) reduces their dominance
+ * while still giving them proportionally more influence.
+ */
+const computeSqrtWeightedRatio = (fileMetrics: { charCount: number; tokenCount: number }[]): number | null => {
+  let weightedRatioSum = 0;
+  let totalWeight = 0;
+  for (const fm of fileMetrics) {
+    if (fm.charCount > 0) {
+      const weight = Math.sqrt(fm.charCount);
+      weightedRatioSum += (fm.tokenCount / fm.charCount) * weight;
+      totalWeight += weight;
+    }
+  }
+  return totalWeight > 0 ? weightedRatioSum / totalWeight : null;
+};
+
+/**
+ * Count tokens for all output parts by falling back to direct token counting.
+ */
+const countAllOutputPartsTokens = async (
+  outputParts: string[],
+  config: RepomixConfigMerged,
+  calculateOutputMetricsFn: typeof calculateOutputMetrics,
+  taskRunner: TaskRunner<TokenCountTask, number>,
+): Promise<number> => {
+  const counts = await Promise.all(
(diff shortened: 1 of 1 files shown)


yamadashy:
## Concern: Accuracy of estimated `totalTokens`

Great optimization — the benchmark numbers are impressive. A few concerns worth discussing before merge:

### 1. `totalTokens` estimation applies file-content ratio to output chars

The sqrt-weighted char:token ratio is derived from **file content**, but `totalTokens` is estimated by applying this ratio to `totalCharacters` — which is the **rendered output** including XML/Markdown tags, headers, directory tree, separators, and git diff/log sections. These structural elements tokenize differently from source code, so the ratio's domain doesn't match. The 0.13% error benchmark may not hold across different output formats (especially XML with heavy tag overhead).

The old code counted tokens on the full output string directly via tiktoken, which was format-agnostic and exact. This was reportedly the most expensive operation (400-800ms), so restoring it would negate much of the speedup — but the tradeoff is worth being explicit about.

### 2. `tokenCountTree` mode no longer counts all files exactly

The old code intentionally calculated token counts for **all files** when `tokenCountTree` was enabled. The new code samples only 50 files and estimates the rest via the ratio. Since `tokenCountTree` is specifically for inspecting per-file token breakdowns, users of this feature likely expect exact values, not estimates. Files with atypical char:token ratios (heavily commented code, minified files, config files) could show noticeably wrong numbers.

Consider keeping the old behavior (all files exact) when `tokenCountTree` is enabled, and applying the sampling optimization only to the default mode.

### 3. Minor items already fixed

The following issues from bot reviews have been addressed in commit a2897ddf:
- Split-output fallback only counting `outputParts[0]` (now counts all parts)
- `totalWeight === 0` returning 0 instead of falling back to direct counting
- Sequential await blocking git metrics (restored `Promise.all` parallelism)
- Duplicated sqrt-weighted ratio loop (extracted `computeSqrtWeightedRatio` helper)

yamadashy:
## Decision: Estimation approach is not acceptable

After further review, we've concluded that **estimated token counts are not acceptable** in any form — neither for `totalTokens` nor for per-file `fileTokenCounts`. Token counts displayed to users and exposed via MCP API must be exact (tiktoken-computed) values.

This means the core optimization of this PR (replacing exact token counting with sampling + ratio estimation) cannot be adopted as-is, since the performance gain comes primarily from skipping exact computation.

### What remains if estimation is removed

| Change | Impact |
|--------|--------|
| Normal mode: 50 → 20 sampled files | Minimal speedup |
| `totalTokens`: ratio estimation | ❌ Must remain exact |
| `tokenCountTree`: 50 files + estimation | ❌ Must remain exact (all files) |

The benchmark improvement (20-27%) is almost entirely attributable to the estimation shortcuts. Removing them leaves negligible gains.

### Suggestion

If you'd like to pursue token counting performance improvements, consider alternative approaches that preserve exact counts, such as:
- Parallelizing tiktoken computation more aggressively (e.g., chunking the output string across workers)
- Caching token counts for unchanged files between runs
- Lazy/streaming token counting overlapped with output generation

Thank you for the thorough benchmarking and the sqrt-weighted ratio analysis — the statistical approach was well-researched. Unfortunately it doesn't fit our requirement for exact token counts.

PR #1409, opened 2026-04-05
perf(core): Reduce worker thread contention for faster pipeline execution

yamadashy:
Add `maxWorkerThreads` option to `WorkerOptions` for explicit thread count capping, then use it to reduce CPU contention when metrics and security worker pools run concurrently during the pipeline overlap phase.

### Changes

- **`processConcurrency.ts`**: Add optional `maxWorkerThreads` to `WorkerOptions`. `getWorkerThreadCount` and `createWorkerPool` respect this cap when provided.
- **`calculateMetrics.ts`**: Cap metrics workers at `processConcurrency - 1` in `createMetricsTaskRunner`. Warmup thread count now matches the actual pool size.
- **`securityCheck.ts`**: Cap security workers at `floor(processConcurrency / 2)`. Security check uses coarse-grained batches (50 files), so fewer workers still provide sufficient parallelism.

### Rationale

During the pipeline overlap phase, the metrics pool (gpt-tokenizer warmup) and security pool run concurrently. On a 4-core machine, this previously created 8 threads (4+4) competing for 4 cores, causing significant context-switching overhead. With the caps, total concurrent threads drop to 5 (3+2), reducing contention while maintaining throughput.

The implementation uses an explicit `maxWorkerThreads` parameter rather than manipulating `numOfTasks`, keeping the API clean and ensuring warmup thread count stays in sync with the actual pool size.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1409" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/core/security/securityCheck.ts
+++ b/src/core/security/securityCheck.ts
@@ -1,6 +1,9 @@
 import pc from 'picocolors';
 import { logger } from '../../shared/logger.js';
-import { getProcessConcurrency, initTaskRunner } from '../../shared/processConcurrency.js';
+import {
+  getProcessConcurrency as defaultGetProcessConcurrency,
+  initTaskRunner,
+} from '../../shared/processConcurrency.js';
 import type { RepomixProgressCallback } from '../../shared/types.js';
 import type { RawFile } from '../file/fileTypes.js';
 import type { GitDiffResult } from '../git/gitDiffHandle.js';
@@ -29,6 +32,7 @@ export const runSecurityCheck = async (
   gitLogResult?: GitLogResult,
   deps = {
     initTaskRunner,
+    getProcessConcurrency: defaultGetProcessConcurrency,
   },
 ): Promise<SuspiciousFileResult[]> => {
   const gitDiffItems: SecurityCheckItem[] = [];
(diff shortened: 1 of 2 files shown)


yamadashy:
Thanks for the thorough review!

**1. macOS regression** — Addressed in 46a0a7b by removing the metrics worker cap (`P-1`). The security cap alone (`min(2, P)`) now shows consistent improvements across all platforms without macOS regression.

**2. Removed `numOfTasks` comment** — Restored in f49a5d06 as a brief inline note.

**3. `getProcessConcurrency` not in `deps`** — Noted. The module mock approach works for now, but could be revisited if testability becomes an issue.

Also added early return for `totalItems === 0` (CodeRabbit nitpick) in the same commit. 🤖

yamadashy:
Addressed the remaining feedback:

**`getProcessConcurrency` injected via `deps`** — Moved from direct module import to the `deps` parameter in `runSecurityCheck`, consistent with how `initTaskRunner` is already injected. Tests updated to pass it explicitly. (7de69232) 🤖

PR #1459, opened 2026-04-13
fix(output): Warn and continue when split-output root entry exceeds size limit

nuthalapativarun:
When using `--split-output` with a limit smaller than a single root directory's output size, Repomix threw a fatal error and exited with no output:

```
✖ Cannot split output: root entry 'src' exceeds max size.
  Part size 27,725,781 bytes > limit 1,048,576 bytes.
```

Since Repomix splits at the root-directory boundary, it cannot split within a single large directory — but aborting entirely leaves the user with no output at all, which is worse than producing an oversized part.

## Changes

- `src/core/output/outputSplit.ts`: Both code paths that previously threw `RepomixError` for an oversized single root entry now emit a `logger.warn` and place the oversized entry in its own output part. The rest of the split continues normally.
- `tests/core/output/outputSplit.test.ts`: Updated the existing "throws error" test to assert the new warn-and-continue behaviour; added a second test covering multiple oversized entries.

## New behaviour

```
⚠ Root entry 'src' exceeds the split size limit (27,725,781 bytes > 1,048,576 bytes). It will be placed in its own output part.

✔ Packing completed successfully!
```

The output is still split into as many parts as possible; the oversized entry just gets its own part. Users who need smaller parts are directed to use `--include` to process subsets of the repository separately.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1459" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


nuthalapativarun:
@coderabbitai review

yamadashy:
Thanks for the PR. Please open an issue to discuss the design first before submitting a PR. Closing this for now.

PR #1475, opened 2026-04-18
perf(config): Defer zod loading to eliminate ~142ms from CLI startup

yamadashy:
## Summary

Move `defaultConfig` and `defaultFilePathMap` to a new `configDefaults.ts` module that has no zod dependency. Remove redundant zod schema validation from `mergeConfigs()` and `buildCliConfig()` — inputs are already validated at their respective boundaries (Commander for CLI, zod in `loadAndValidateConfig` for config files). Lazy-load `configSchema.ts` (which imports zod) only when a config file actually needs validation.

Add speculative preload in `cliRun.ts`: when common config files (json/json5/jsonc) are detected via `fs.access`, start loading `configSchema.js` concurrently with the `defaultAction` import chain so zod is ready by the time validation runs. This prevents regression for repos with config files.

## Why it works

Before: zod (~145ms) was always loaded as a static dependency of `configSchema.ts`, blocking the module import phase on every CLI invocation regardless of whether validation was needed.

After: zod is only loaded when a config file exists and needs validation. When no config file is present (common for many users), zod is never loaded.

## Benchmark

Original measurement (interleaved A/B, 10 pairs):

Without config file:
| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Median | 1529ms | 1387ms | **-142ms (-9.3%)** |
| Mean   | 1533ms | 1386ms | **-147ms (-9.6%)** |

With config file (no regression):
| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Median | 1660ms | 1674ms | ~0ms (neutral) |

The ~142ms saving matches the measured zod module load time (145ms).

Note: Independent re-measurement on a different machine showed varied results (round 1 +2.5ms, round 2 -19.6ms with high noise). Effect may depend on hardware and Node.js compile cache state.

## Checklist

- [x] Run \`npm run test\` (1114 passed; one redundant zod validation test removed)
- [x] Run \`npm run lint\`

---

<a href="https://app.devin.ai/review/yamadashy/repomix/pull/1475" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/cli/actions/defaultAction.ts
+++ b/src/cli/actions/defaultAction.ts
@@ -1,16 +1,11 @@
 import path from 'node:path';
+import { OUTPUT_STYLES, type RepomixOutputStyle } from '../../config/configDefaults.js';
 import { loadFileConfig, mergeConfigs } from '../../config/configLoad.js';
-import {
-  type RepomixConfigCli,
-  type RepomixConfigFile,
-  type RepomixConfigMerged,
-  type RepomixOutputStyle,
-  repomixConfigCliSchema,
-} from '../../config/configSchema.js';
+import type { RepomixConfigCli, RepomixConfigFile, RepomixConfigMerged } from '../../config/configSchema.js';
 import { readFilePathsFromStdin } from '../../core/file/fileStdin.js';
 import { type PackResult, pack } from '../../core/packager.js';
 import { generateDefaultSkillName } from '../../core/skill/skillUtils.js';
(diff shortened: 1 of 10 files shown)


yamadashy:
Closing in favor of #NEW which preserves the public API's validation behavior (only the function signatures change from sync to async). See replacement PR.

yamadashy:
Replaced by #1484 — keeps full validation behavior of the public API (only signature becomes async). Independently re-measured: -15.7ms (-4.6%) average improvement when no config file is present, neutral when present.

PR #1576, opened 2026-05-16
perf(core): Overlap metrics warm-up and output module load with searchFiles

yamadashy:
## Summary

Reorders the pack startup so two pre-existing latencies stop sitting on the critical path between `searchFiles` and `produceOutput`. Section 1 of PR #1548 (token-count disk cache) is already on `main`; this extends it.

### What changed

Three coupled changes in `packager.ts` / `calculateMetrics.ts`:

- **Pre-init the metrics worker pool BEFORE `searchFiles`.** Each metrics worker pays ~200–285ms to parse gpt-tokenizer's BPE table. Until now this overlap window started AFTER search; moving it earlier absorbs the ~100–150ms search phase into the parallel pipeline. Pool sizing no longer depends on file count — capped at `METRICS_PREWARM_THREAD_CAP = 3` so the pool can be created without waiting on search.
- **Lazy-load `produceOutput`** through the dynamic-import deps pattern already used for `packSkill`. The output-style + Handlebars chain now resolves in parallel with `searchFiles` / collect rather than blocking packager.ts entry.
- **Drop the `await metricsWarmupPromise` / `await tokenCacheLoadPromise` barriers** before `produceOutput`. Neither `produceOutput` nor file collection depend on those promises; the awaits move into the metrics branch of the trailing `Promise.all`, where they are actually needed.

The `try` / `finally` block is reshaped so `metricsTaskRunner.cleanup()` still fires if `searchFiles` itself throws. A new packager.test pins this leak path.

### What was dropped from the original plan

An earlier revision of this PR also bundled a **security pre-screen** (`QUICK_SECRET_SCREEN` regex in `securityCheckWorker.ts`) that skipped `lintSource` for files that obviously could not contain a secret. It was removed because it weakened the safety invariant: instead of "every file content is scanned by secretlint", repomix would have scanned "every file content that matches a hand-maintained marker regex". Any future secretlint upgrade that adds a rule (or changes a token shape) without a matching marker update would silently drop real secrets. The performance gain (≈3–4% extra wall-clock on the full repo) was not worth the regression in security guarantee — kept only the strictly-safe pipeline reordering.

## Benchmarks (warm cache, paired, hyperfine `--runs 25`, repomix self-pack on darwin/arm64)

| scope | files | BEFORE (main) | AFTER (this PR) | Δ |
|---|---|---|---|---|
| **default (whole repo)** | ~1050 | 812.8 ± 27.4 ms | 737.6 ± 9.0 ms | **−75.2 ms / −9.3%** |
| medium (`src/**,tests/**`) | ~470 | 304.4 ± 14.6 ms | 294.6 ± 5.1 ms | −9.8 ms / −3.2% |

- Output `repomix --quiet` is **byte-identical** between BEFORE and AFTER at every scope (verified by MD5).
- User CPU on the full repo drops from 2292 ms → 1149 ms (≈ half), reflecting the new `METRICS_PREWARM_THREAD_CAP = 3` cap on the metrics pool. On hosts with >3 vCPUs and very large file counts this is a deliberate trade — the metrics phase now overlaps `produceOutput` and is no longer on the critical path after the per-file token-count cache landed.

## Notes

- Branched from `main`, not cherry-picked. The idea originates in PR #1548 but was rewritten against current code (token-count cache + UTF-8 ordering + 1-pass globby are already on `main`).
- All landed regression specs from PR #1575 still pass: gitignore, dot-ignore, empty-directory, transform-order, binary-detection, multi-root, security-scan.

## Checklist

- [x] Run `npm run test` — 1304 → 1305 passing (+1 new `searchFiles` cleanup leak test)
- [x] Run `npm run lint` — clean (3 pre-existing warnings only)

### DIFF
--- a/src/core/metrics/calculateMetrics.ts
+++ b/src/core/metrics/calculateMetrics.ts
@@ -1,6 +1,6 @@
 import type { RepomixConfigMerged } from '../../config/configSchema.js';
 import { logger } from '../../shared/logger.js';
-import { getWorkerThreadCount, initTaskRunner } from '../../shared/processConcurrency.js';
+import { getProcessConcurrency, initTaskRunner } from '../../shared/processConcurrency.js';
 import type { RepomixProgressCallback } from '../../shared/types.js';
 import type { ProcessedFile } from '../file/fileTypes.js';
 import type { GitDiffResult } from '../git/gitDiffHandle.js';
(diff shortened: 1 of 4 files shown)


yamadashy:
Superseded by #1580. Bisecting the 4 changes in this PR (early init / cap / lazy import / drop blocking awaits) showed that the cap on metrics workers explained essentially all of the wall-clock win. #1580 isolates that as a sub-millisecond `existsSync` probe on the token-count cache file with no cold-cache regression, so the rest of this bundle is not worth landing.

PR #1577, opened 2026-05-17
perf(core): Overlap metrics warm-up with searchFiles (#1 + #4 only)

yamadashy:
## Summary

Companion / alternative to PR #1576. That PR bundles four coupled changes; this one isolates just the two that the analysis pointed to as the most likely contributors to the CI benchmark gains:

1. **Move `createMetricsTaskRunner` to BEFORE `searchFiles`.** The ~200-285ms-per-worker gpt-tokenizer BPE table parse now overlaps the search phase too, not just collect/security/process. Pool sizing switches from file-count-based (`min(cpu, ceil(N/100))`) to host-concurrency-based (`getProcessConcurrency()`) so it can be constructed without waiting on `searchFiles`.

2. **Drop the `await metricsWarmupPromise` / `await tokenCacheLoadPromise` barriers** that sat just before `produceOutput`. They move inside the metrics branch of the trailing `Promise.all`, where they are actually needed.

Deliberately **not included** here:
- The hard `METRICS_PREWARM_THREAD_CAP = 3` on the pool (from PR #1576). This branch keeps the pool sized by `getProcessConcurrency()` so very-large-repo / many-vCPU hosts retain the same metrics parallelism as `main`.
- The lazy `produceOutput` import (from PR #1576).

The goal is to measure CI benchmarks for this minimal variant and compare against PR #1576, so we know which knob each contribution is buying.

## Safety nets

- `try`/`finally` block expands upward to wrap `searchFiles` so a search failure still triggers `metricsTaskRunner.cleanup()`. New packager test pins the leak path.
- `createMetricsTaskRunner` catches its own synchronous post-pool failures so a half-constructed runner cannot escape.
- `outputForMetricsPromise` carries a no-op `.catch` to avoid an unhandled-rejection window if `produceOutput` rejects while the metrics branch is still awaiting the warm-up promise.

## Checklist

- [x] Run `npm run test` — 1304 → 1305 passing (+1 new `searchFiles` cleanup leak test)
- [x] Run `npm run lint` — clean (3 pre-existing warnings only)

### DIFF
--- a/src/core/metrics/calculateMetrics.ts
+++ b/src/core/metrics/calculateMetrics.ts
@@ -1,6 +1,6 @@
 import type { RepomixConfigMerged } from '../../config/configSchema.js';
 import { logger } from '../../shared/logger.js';
-import { getWorkerThreadCount, initTaskRunner } from '../../shared/processConcurrency.js';
+import { getProcessConcurrency, initTaskRunner } from '../../shared/processConcurrency.js';
 import type { RepomixProgressCallback } from '../../shared/types.js';
 import type { ProcessedFile } from '../file/fileTypes.js';
 import type { GitDiffResult } from '../git/gitDiffHandle.js';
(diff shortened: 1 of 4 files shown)


yamadashy:
Closing. Isolating early-init + drop-blocking-awaits (#1 + #4 from #1576) showed no benefit and a slight regression on Windows CI (+2.6%); the win in #1576 came from the metrics-pool prewarm reduction (#2), which is implemented in #1580.

PR #1590, opened 2026-05-22
fix(website): remove invalid Turnstile size 'invisible', hide container via CSS

### LINKED ISSUE #1551: Cloudflare Turnstile reports invalid 'size'
### Description

1. Open up repomix.com
2. Try to use the service. e.g. upload a .zip file
3. Click "Pack"

Expected: packing process completes succesfully
Actual: packing is interrupted by a TurnstileException, which prevents the app from processing 
Was reproducible for me every time I tried.

Image

Image

Transcript: `Uncaught TurnstileError: [Cloudflare Turnstile] Invalid value for parameter "size", expected "compact", "flexible", or "normal", got "invisible".`

### Usage Context

repomix.com

### Repomix Version

_No response_

### Node.js Version

_No response_

### Environment

Chrome, v. 147.0.7727.139, Windows 11 26200.8246

serhiizghama:
## Problem

Cloudflare Turnstile throws a `TurnstileError` when `size: 'invisible'` is passed to `render()`:

```
Uncaught TurnstileError: [Cloudflare Turnstile] Invalid value for parameter "size",
expected "compact", "flexible", or "normal", got "invisible".
```

This breaks the Pack form on repomix.com for affected users (reported in #1551).

The valid values for the `size` parameter are `'normal'`, `'compact'`, and `'flexible'`. The value `'invisible'` was never a documented Turnstile option and is rejected by recent Turnstile script versions.

## Solution

1. **`useTurnstile.ts`** — remove `size: 'invisible'` from the `render()` options. The non-interactive behaviour is already provided by `execution: 'execute'`, which defers the challenge until `mintToken()` is called.

2. **`useTurnstileScript.ts`** — update the `TurnstileRenderOptions.size` type to match the actual API (`'normal' | 'compact' | 'flexible'`).

3. **`TryIt.vue`** — add a `.turnstile-container` CSS rule (`position: absolute; width/height: 0; overflow: hidden; visibility: hidden`) so the widget does not consume layout space or appear visually when rendered without a size.

## Testing

- `npm run lint` — no errors
- `npm run test` — all 1312 tests pass
- Verified the render options no longer include `size: 'invisible'`

Fixes #1551

### DIFF
--- a/website/client/components/Home/TryIt.vue
+++ b/website/client/components/Home/TryIt.vue
@@ -478,4 +478,16 @@ onMounted(() => {
   border-color: #333 transparent transparent transparent;
 }
 
+/* The Turnstile widget is executed programmatically via execution: 'execute'.
+   The container must remain in the DOM (removing it prevents widget rendering)
+   but should not affect page layout or be visible. */
+.turnstile-container {
+  position: absolute;
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  visibility: hidden;
+  pointer-events: none;
+}
+
 </style>
--- a/website/client/composables/useTurnstile.ts
+++ b/website/client/composables/useTurnstile.ts
@@ -87,7 +87,6 @@ export function useTurnstile() {
       if (!widgetId.value) {
         widgetId.value = turnstile.render(el, {
           sitekey: siteKey,
-          size: 'invisible',
           action: 'pack',
           execution: 'execute',
           callback: (token: string) => {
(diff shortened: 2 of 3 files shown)


yamadashy (approved):

yamadashy:
Hi @serhiizghama !
Thanks for the fix!
We didn't notice the implementation issue. Merging as-is.

The Cloudflare console config on our side might also be off, so we'll take a look as a follow-up.

PR #1631, opened 2026-06-09
fix(cli): copy split output files after remote packing

Samsen879:
## Summary

I noticed that remote mode does not handle `--split-output` correctly right now.

When Repomix packs a remote repository, it works in a temporary directory first and then copies the generated output back to the current directory. That is fine for normal output, because there is only one output file.

With `--split-output`, the actual files are numbered files like:

- `repomix-output.1.xml`
- `repomix-output.2.xml`

The packer already returns those filenames in `packResult.outputFiles`, but `runRemoteAction` was still trying to copy only `config.output.filePath`. In split mode that base file may not exist, so the remote run can fail while the actual split files stay in the temporary directory.

This PR makes remote mode copy the returned `outputFiles` when they are present, and keeps the old single-file behavior as the fallback.

## Tests

- `npx vitest run tests/cli/actions/remoteAction.test.ts`
- `npx vitest run tests/cli/actions/remoteAction.test.ts tests/core/packager/splitOutput.test.ts tests/core/packager/produceOutput.test.ts`
- `npx biome check src/cli/actions/remoteAction.ts tests/cli/actions/remoteAction.test.ts`
- `npm run lint-ts`

### DIFF
--- a/src/cli/actions/remoteAction.ts
+++ b/src/cli/actions/remoteAction.ts
@@ -152,7 +152,10 @@ export const runRemoteAction = async (
     // For skill generation, the skill is already written directly to the target directory
     // (either via --skill-output path or via promptSkillLocation which uses process.cwd())
     if (!cliOptions.stdout && result.config.skillGenerate === undefined) {
-      await copyOutputToCurrentDirectory(tempDirPath, process.cwd(), result.config.output.filePath);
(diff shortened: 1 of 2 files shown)


yamadashy (approved):

yamadashy:
Hi, @Samsen879 !
Thanks for the PR! The code looks good to me, so I will merge it. 🙏

PR #1644, opened 2026-06-15
Add watch mode for local rebuilds

### LINKED ISSUE #1429: file watch option
### Description

is it possible to have a `--watch` option so repomix can rebuild it self instead of doing so manually.

Samsen879:
Closes #1429.

## Summary

This adds a local `--watch` mode for rebuilding Repomix output after file changes, so the CLI does not need to be restarted during edit loops.

The watcher runs an initial pack, then debounces change events before running the normal pack flow again. It also avoids self-triggering on generated output files and skips the browser prompt while watch mode is still running.

A few modes are rejected up front because they do not fit a long-lived rebuild loop: remote repositories, stdin, stdout, clipboard copy, and skill generation.

## Testing

- `npm run lint`
- `npm run build`
- `npm test -- --run`
- manual watch smoke test: started `repomix --watch`, edited a local file, and confirmed the output was rebuilt

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,12 @@ using [glob patterns](https://github.com/mrmlnc/fast-glob?tab=readme-ov-file#pat
 repomix --include "src/**/*.ts,**/*.md"
 ```
 
+To rebuild the output automatically when local files change:
+
+```bash
+repomix --watch
+```
+
 To exclude specific files or directories:
 
 ```bash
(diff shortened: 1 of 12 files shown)


yamadashy:
Thank you so much for working on this, @Samsen879! 🙏

The `--watch` feature has since landed via #1643 and shipped in **v1.15.0**, so this PR now overlaps with what is already on `main` (which is why it shows conflicts). I am going to close this one as superseded, but I really appreciate you taking the time to design and implement it.

PR #1646, opened 2026-06-15
Add cwd-relative output path style

### LINKED ISSUE #1424: Add more control how file paths in output are formatted
### Description

Currently file paths in output are relative to the analysis folder, e.g. if you do `repomix .` in project folder, and it has `src/main.py`, you'll see:
```<file path="src/main.py">```

Sometimes you have an extension of project's source code that's in a different folder from current working directory, e.g. ```../core-library/```. So then you can do ```repomix ../core-library/```. E.g. this one has file `src/core.py`, and in output you'll see:
```<file path="src/core.py">```

Suggestion: add optional modes how these file paths should be formatted in the output. E.g. in this case it would be preferrable to instead have ```<file path="../core-library/src/core.py">``` in the output, which is relative to the current working directory.

I'm trying an approach where if it's a smaller project, I ask agent to run repomix and get all project-related source code into context. But this pathing makes it forget that not all files are under the current working directory, so it wastes time searching for the file in current working directory, then realizes it's not there, and goes to the correct folder.

Samsen879:
## Summary

- Add `output.filePathStyle` with the existing `target-relative` behavior as the default.
- Add `cwd-relative` support so file paths are rendered relative to the command's current working directory.
- Add `--output-file-path-style` for the CLI and update the docs/schema.

Closes #1424

## Testing

- `npm test -- --run`
- `npm run build`
- `npm run lint`
- Manual smoke test for `repomix ../core-library --output-file-path-style cwd-relative`

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,7 @@ Instruction
 |--------|-------------|
 | `-o, --output <file>` | Output file path (default: `repomix-output.xml`, use `"-"` for stdout) |
 | `--style <style>` | Output format: `xml`, `markdown`, `json`, or `plain` (default: `xml`) |
+| `--output-file-path-style <style>` | How file paths are shown in output: `target-relative` or `cwd-relative` (default: `target-relative`) |
 | `--parsable-style` | Escape special characters to ensure valid XML/Markdown (needed when output contains code that breaks formatting) |
 | `--compress` | Extract essential code structure (classes, functions, interfaces) using Tree-sitter parsing |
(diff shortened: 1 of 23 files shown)


yamadashy (approved):

yamadashy:
Thanks for the PR, @Samsen879! 🙏

The code looks good. I pushed a couple of small follow-ups and rebased on main, so CI is green now. I'll go ahead and merge.

Thanks again!

PR #1647, opened 2026-06-15
Address watch mode review feedback: reject `--watch` + `--copy`, validate merged config, tests & docs

PAMulligan:
Follow-up to #1643, built on top of `feat/file-watch-option` (`b920430a`). `src/cli/actions/watch/watchIgnore.ts` is intentionally left untouched.

Each change maps to the open review items:

### 1. `--watch` + `--copy` / config-file conflicts
`--copy` is now treated as a **hard conflict** with `--watch` (not copy-on-initial-pack), consistent with the other hard conflicts:
- `validateWatchOptions` (cliRun) rejects `--watch --copy` on the CLI.
- `runWatchAction` routes the merged config through the shared `validateConflictingOptions(config)` and rejects config-file-driven `copyToClipboard`, stdout (`output: "-"` via `filePath === '-'`), and `skillGenerate` — so the watch route validates identically to the default route, including options that `validateWatchOptions` (CLI-flags-only) can't see.

### 2. Shutdown-path test gaps
Added tests for the production `process.on('SIGINT'/'SIGTERM')` + `removeListener` path (previously never exercised because the flow tests always inject an `AbortSignal`), the watcher `error` handler, the rebuild-failure path (`pack` rejects → `logger.error('Watch rebuild failed')`), double-Ctrl+C cleanup idempotency, the already-aborted early-return guard, `watcher.close()` failure during shutdown, and the merged-config stdout/copy/skill rejections. `watchAction.ts` line coverage is now ~94% (was ~90%).

### 3. Docs (English only)
Added a **Watch Mode Options** section to `website/client/src/en/guide/command-line-options.md` and an incompatibility note to the README. Other-language translations deferred, as discussed.

### 4. Minor cleanups
- `Rebuilt at` timestamp uses the system locale (`undefined`) instead of hardcoded `en-GB`.
- Extracted the 300 ms rebuild debounce and 100 ms `awaitWriteFinish` stability threshold into named constants.
- Added a comment documenting the EMFILE behavior (after a fatal watcher error the process stays alive but is silently no longer watching).
- Left `monitor`/`live`/`auto` as `semanticSuggestionMap` suggestions (not aliases), as agreed.

### Note on the docs list
Since item 1 makes `--copy` a hard conflict, I included `--copy` in the docs incompatibility list (item 3's list didn't mention it) so the docs stay accurate. Happy to drop it if you'd prefer.

---
Full `npm run test` (**1390 passed**) and lint (biome / oxlint / tsgo / secretlint) are green. Commits split as `fix(cli)`, `test(cli)`, `docs`, `refactor(cli)`. Not merging — leaving this for your review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,6 +692,8 @@ Instruction
 #### Watch Mode
 - `-w, --watch`: Watch for file changes and automatically re-pack. Debounces rapid changes (300ms) and logs a timestamp on each rebuild. Stop with `Ctrl+C`.
 
+  Watch mode only works with local directories, so it cannot be combined with `--remote`, a positional remote repository URL, `--stdout`, `--stdin`, `--split-output`, `--skill-generate`, or `--copy` (whether set on the command line or in your config file).
+
 #### Examples
 
 ```bash
--- a/src/cli/actions/defaultAction.ts
+++ b/src/cli/actions/defaultAction.ts
@@ -383,11 +383,8 @@ export const buildCliConfig = (options: CliOptions): RepomixConfigCli => {
 /**
  * Validates that conflicting CLI options are not used together.
  * Throws RepomixError if incompatible options are detected.
- *
- * Exported so the watch action can apply the same validation to its merged
- * config, keeping the watch route consistent with the default route.
  */
-export const validateConflictingOptions = (config: RepomixConfigMerged): void => {
+const validateConflictingOptions = (config: RepomixConfigMerged): void => {
   const isStdoutMode = config.output.stdout || config.output.filePath === '-';
 
   // Define option states for conflict checking
--- a/src/cli/actions/defaultAction.ts
+++ b/src/cli/actions/defaultAction.ts
@@ -383,8 +383,11 @@ export const buildCliConfig = (options: CliOptions): RepomixConfigCli => {
 /**
  * Validates that conflicting CLI options are not used together.
  * Throws RepomixError if incompatible options are detected.
+ *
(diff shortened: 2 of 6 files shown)


PAMulligan:
@coderabbitai review

PAMulligan:
Thanks @gemini-code-assist — addressed both points in ce858161 and fea83bf6:

**Timestamp** — switched to `new Date().toTimeString().split(' ')[0]` for a guaranteed locale-independent `HH:MM:SS`. You're right that `toLocaleTimeString(undefined, …)` follows the *system* locale and can render non-ASCII digits / AM-PM separators, which defeats the portability goal.

**Redundant `validateConflictingOptions(config)`** — removed the call (and its dedicated test), and reverted the now-unneeded `export` in `defaultAction.ts`. Watch mode forbids `splitOutput` / `stdout` / `skillGenerate` / `copyToClipboard` individually, so the shared validator was redundant and could surface a confusing *pairwise* message (e.g. "Skill output is a directory and cannot be copied to clipboard") instead of a clear "--watch cannot be used with X". The per-option watch checks still cover every config-file case, including `output: "-"` (stdout) and `skillGenerate`.

@yamadashy heads-up: both tweaks deviate slightly from the plan we agreed (which specified `toLocaleTimeString(undefined)` and routing through `validateConflictingOptions`). The bot's reasoning looks sound on both, but happy to restore either if you'd rather keep the original approach.

yamadashy:
@PAMulligan 
This looks great, merging now. 
Thanks for the thorough follow-up!

PR #1653, opened 2026-06-17
feat(output): add per-file inclusion levels (output.patterns)

### LINKED ISSUE #608: Files "inclusion level"
Create a way to provide the detalization level for packed files.

There are 3 levels:

1. **Full content** - for a set of the most important files.
2. **Compress** - for the less important ones, which still may be relevant.
3. **Directory-only** - mention in the directory structure.

One can think of it as nested sets which describe not just the files, but the level of their detalization:

[ File in directory structure] < [ Compressed file ] < [ Full file ] 

Use case: we need to improve build system.
1. Full files: for all configs and build files.
2. Directory structure: full file structure.

Use case: we need to work on a module residing inside a certain folder:
1. Full files: the whole module, main project files, configuration
2. Compressed files: other modules which somewhat contribute to context, but details aren't important.

Then we'll be able to create packing scenarios with various such settings and invoke them via MCP.

The simplest syntax would be 3 settings, e.g.

```js
files: {
  content: [...globs...],  // full content
  compress: [...globs...], // compressed
},
directory: [...globs...] // only list in the file structure
```

1. If a file matches a glob in `content` - full content is included.
2. Otherwise, if it matches a glob in `compress` - it gets compressed

So we can have like `compress: **/*.ts, content: my.ts` - all ts will be compressed, but my.ts will be included fully.

Also, all mentioned files are included into the directory structure, together with `directory` globs.
The `directory` globs has precedence over all ignores.

Here, the `directory` setting is the least important one, but it still may matter for huge projects with many files and folders. I understand we already have some ignore settings, perhaps they can replace `directory` or somehow make it into the mix =)

PAMulligan:
## Summary

Implements per-file **inclusion levels** for the packed output, as proposed and designed in #608. A new optional `output.patterns` array in the config lets you control each file's detail level on a per-glob basis, with three levels:

- **Full content** (default) — current behavior.
- **Compressed** (`compress: true`) — runs the file through the existing `--compress` / Tree-sitter pipeline.
- **Directory-structure-only** (`directoryStructureOnly: true`) — the file is listed in the directory structure, but its content block is omitted.

```json5
{
  "output": {
    "compress": false,            // global default acts as the catch-all
    "patterns": [
      { "pattern": "docs/**/*", "compress": true },
      { "pattern": "website/**/*", "directoryStructureOnly": true }
    ]
  }
}
```

This follows the `output.patterns` design agreed in the issue thread.

### Semantics

- Patterns are evaluated in array order; the **first matching pattern wins** for a file.
- A matched pattern's flags **override** the global `output.compress` for that file. `directoryStructureOnly` takes precedence over `compress`. A matched pattern with no flags forces full content (useful for whitelisting files out of a global `compress`).
- If no pattern matches, the existing global behavior applies (full content, or compressed when `output.compress` is `true`).
- **Backward compatible** — configs without `patterns` behave exactly as before.
- `output.patterns` only affects *how* a collected file is rendered; `include`/`ignore` still decide *whether* a file is collected.

### Implementation notes

- New pure helper `resolveFileLevel(filePath, output)` → `'full' | 'compress' | 'directory-only'`, centralising the first-match-wins logic. Globs are matched with `minimatch({ dot: true })`, the same way `fileSearch` matches ignore patterns — no new dependency.
- `processFiles` resolves each file's level once: directory-only files are dropped from the content output, while their paths still feed the directory tree (the tree is built from the file-search results, not from `processedFiles` — the same mechanism the existing suspicious-file filter relies on). This covers the empty/directory edge case raised in the thread: a directory whose files are all directory-only still appears in the structure.
- Compression is routed per file, so a `compress` pattern works even when global `compress` is off. The `showLineNumbers` suppression and the "content has been compressed" header note are likewise per-file/per-pattern aware.
- `output.patterns` is added to the config schema (valibot) and validated strictly (`pattern` required string; `compress`/`directoryStructureOnly` optional booleans).

### Tests

- `tests/core/file/fileLevelResolve.test.ts` — resolver units: ordering, three levels, no-patterns backward-compat, override, dotfiles, posix path matching.
- `tests/core/file/fileProcess.inclusionLevel.test.ts` — pipeline wiring: directory-only exclusion, per-pattern compress with global off, global-compress override, line-number suppression.
- `tests/core/output/inclusionLevelOutput.test.ts` — end-to-end: a directory-only file (and a directory whose files are all directory-only) appears in the tree with content omitted.
- Additions to `tests/config/configSchema.test.ts`, `tests/config/configLoad.test.ts`, `tests/core/output/outputStyleDecorate.test.ts`.
- Full suite: 1422 passed, 10 skipped, 0 failed. `npm run lint` clean (biome, oxlint, tsgo, secretlint).

### Notes / scope

- Config-file only; no CLI flag (per the agreed design).
- English docs updated (`README.md`, `website/client/src/en/guide/configuration.md`); other locales can be translated separately.
- The versioned JSON schemas under `website/client/src/public/schemas/` are release-time artifacts generated from the file schema, so they're left untouched here and will pick up `patterns` on the next `website-generate-schema` run.
- Intentionally out of scope (later ideas in the thread): MCP packing scenarios, and per-pattern control of other options like `removeComments` (the schema is structured to allow adding these later).

Resolves #608.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/core/file/fileProcess.ts
+++ b/src/core/file/fileProcess.ts
@@ -3,6 +3,7 @@ import type { RepomixConfigMerged } from '../../config/configSchema.js';
 import { logger } from '../../shared/logger.js';
 import { initTaskRunner } from '../../shared/processConcurrency.js';
 import type { RepomixProgressCallback } from '../../shared/types.js';
+import { resolveFileLevel } from './fileLevelResolve.js';
 import { type FileManipulator, getFileManipulator } from './fileManipulate.js';
 import type { ProcessedFile, RawFile } from './fileTypes.js';
 import { truncateBase64Content } from './truncateBase64.js';
@@ -43,7 +44,10 @@ export const applyLightweightTransforms = (
 
     content = content.trim();
 
-    if (config.output.showLineNumbers && !config.output.compress) {
+    // Line numbers are suppressed for compressed files (their content is
+    // restructured signatures, not line-faithful source), matching the global
+    // compress behavior on a per-file basis.
+    if (config.output.showLineNumbers && resolveFileLevel(file.path, config.output) !== 'compress') {
       const lines = content.split('\n');
       const padding = lines.length.toString().length;
       const numberedLines = lines.map((line, idx) => `${(idx + 1).toString().padStart(padding)}: ${line}`);
@@ -84,21 +88,32 @@ export const processFiles = async (
   const startTime = process.hrtime.bigint();
   let files: ProcessedFile[];
 
-  // Only compress (tree-sitter) and removeComments (AST manipulation) justify worker thread overhead
-  const useWorkers = config.output.compress || config.output.removeComments;
+  // Resolve each file's inclusion level once. Files at the 'directory-only'
+  // level are dropped from the content output entirely — their paths still
+  // appear in the directory structure, which is built from the search results
+  // rather than from processedFiles.
+  const leveledFiles = rawFiles
+    .map((rawFile) => ({ rawFile, level: resolveFileLevel(rawFile.path, config.output) }))
+    .filter((entry) => entry.level !== 'directory-only');
+
+  // Only compress (tree-sitter) and removeComments (AST manipulation) justify worker thread overhead.
+  // Compression can be requested globally or per-file via output.patterns, so check the resolved
+  // levels rather than the global compress flag alone.
+  const needsCompression = leveledFiles.some((entry) => entry.level === 'compress');
+  const useWorkers = needsCompression || config.output.removeComments;
 
   if (useWorkers) {
     // Phase 1: Heavy processing via workers (removeComments, compress)
-    logger.trace(`Starting file processing for ${rawFiles.length} files using worker pool`);
+    logger.trace(`Starting file processing for ${leveledFiles.length} files using worker pool`);
 
     const taskRunner = deps.initTaskRunner<FileProcessTask, ProcessedFile>({
-      numOfTasks: rawFiles.length,
+      numOfTasks: leveledFiles.length,
       workerType: 'fileProcess',
       runtime: 'worker_threads',
     });
 
-    const tasks = rawFiles.map(
-      (rawFile) =>
+    const tasks = leveledFiles.map(
+      ({ rawFile }) =>
         ({
           rawFile,
           config,
@@ -130,8 +145,8 @@ export const processFiles = async (
     files = applyLightweightTransforms(files, config, () => {}, deps);
   } else {
     // No heavy processing needed - apply lightweight transforms directly
-    logger.trace(`Starting file processing for ${rawFiles.length} files in main thread (lightweight mode)`);
-    const inputFiles = rawFiles.map((rawFile) => ({ path: rawFile.path, content: rawFile.content }));
+    logger.trace(`Starting file processing for ${leveledFiles.length} files in main thread (lightweight mode)`);
+    const inputFiles = leveledFiles.map(({ rawFile }) => ({ path: rawFile.path, content: rawFile.content }));
     files = applyLightweightTransforms(inputFiles, config, progressCallback, deps);
   }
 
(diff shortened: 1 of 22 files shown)


PAMulligan:
@coderabbitai review

yamadashy:
Thanks for the PR, and for following the `output.patterns` design from #608 so closely. The implementation looks good to me.

It currently conflicts with `main` though. Could you rebase onto the latest `main`? Once that is done and CI is green, I will get it merged.

Thanks again!

PAMulligan:
The only failing check here is **Test coverage**, and it's failing inside the Codecov upload step's own integrity check, not on this change:

```
gpg: no valid OpenPGP data found … Total number processed: 0
gpg: Can't check signature: No public key
==> Could not verify signature ... Exiting...
```

Codecov's CLI couldn't import its signing key, so with `fail_ci_if_error: true` the job aborts before any coverage is uploaded. Every other check (tests across all OS/Node, build-and-run, lint, CodeQL) passes and `main` is currently green on coverage, so this looks like a transient Codecov-side flake. Could you re-run the **Test coverage** job when you get a chance? 🙏

yamadashy:
@PAMulligan 
Thanks!
No worries about the Codecov failure.
The real blocker is the conflict with `main`, so could you rebase onto the latest `main`? Once it is conflict free I will get it merged. 🙏

PAMulligan:
@yamadashy Rebased onto the latest `main` and resolved the conflict — the PR now shows as mergeable. ✅

The only conflict was in `tests/config/configLoad.test.ts`: your `output.filePathStyle` tests and my `output.patterns` test had been added at the same spot, so I kept both. The four commits are otherwise unchanged, and everything is green locally — build, lint/type-check, and the full Vitest suite (1456 passed).

Heads-up: the CI workflows for the new commit are sitting in **"action_required"**, waiting on your approval to run (the fork-PR gate). Once you approve them they should go green — give or take the Codecov upload flake you already waved off. Ready to merge whenever you are. Thanks again! 🙏

src/core/file/fileProcess.ts
```diff
@@ -84,24 +93,40 @@ export const processFiles = async (
...
-  // Only compress (tree-sitter) and removeComments (AST manipulation) justify worker thread overhead
-  const useWorkers = config.output.compress || config.output.removeComments;
+  // Resolve each file's inclusion level once and cache it by path. Files at the
+  // 'directory-only' level are dropped from the content output entirely — their
+  // paths still appear in the directory structure, which is built from the search
+  // results rather than from processedFiles. The cached levels are reused by
+  // applyLightweightTransforms so the glob matching is not repeated.
+  const fileLevels = new Map<string, FileInclusionLevel>();
+  const leveledFiles = rawFiles.flatMap((rawFile) => {
+    const level = resolveFileLevel(rawFile.path, config.output);
```
yamadashy: `rawFile.path` is already the **display path** here (`packager.ts` rewrites it via `buildFileDisplayPath` before `processFiles`), so `output.patterns` matches the display path, not the per-root-relative path `include`/`ignore` use, despite the docs and the `resolveFileLevel` comment saying they match the same way. With multiple roots a pattern then needs the root-label prefix shown in the directory structure (e.g. `packages/app/src/**` instead of `src/**`), and `output.filePathStyle: cwd-relative` shifts it again.

Intended? I think it should match the per-root-relative path like `include`/`ignore`: resolve the level before the display transform in `packager.ts` and thread it here.

yamadashy (commented):
@PAMulligan Thanks for the rebase! One thing on the `output.patterns` matching basis before merge (see inline).

src/core/file/fileProcess.ts
yamadashy: Thanks, `7768f1d2` addresses this exactly: the level is now resolved against the per-root-relative path in `packager.ts` before the display transform and threaded via `RawFile.level`. Looks good 👍

yamadashy:
@PAMulligan 
Thanks again for all the work on this, and for the careful follow-ups!

The code looks good and CI is green.
Merging now 🎉

yamadashy (approved):

PR #1683, opened 2026-06-29
fix(core): strip comments from .mjs/.cjs/.mts/.cts files

serhiizghama:
`--remove-comments` and `--compress` silently leave comments intact in `.mjs`, `.cjs`, `.mts` and `.cts` files, while stripping them from `.js`/`.ts`. They were just missing from the manipulators map in `fileManipulate.ts`.

These are ESM/CommonJS variants of the same languages, and the rest of the codebase already treats them that way — `treeSitter/languageConfig.ts` lists `cjs`/`mjs`/`mts`/`cts` under JavaScript/TypeScript, and the skill stats label `.mjs`/`.cjs` as JavaScript. So a `.mjs` file gets tree-sitter compressed but its comments are never stripped, which is inconsistent. This maps the four to the same `javascript` strip language `.js`/`.ts` already use (TS syntax strips fine under it — the existing `.ts` test relies on that).

Added a test that resolves a manipulator for each extension and strips comments from an `eslint.config.mjs`-style file.


yamadashy:
Thanks for the PR, @serhiizghama! 🙏

The code looks good to me.
I pushed a small follow-up commit to also cover `.mjsx`/`.mtsx` for full parity with `languageConfig`.

Merging now, thanks again! 🎉

yamadashy (approved):

PR #1710, opened 2026-07-09
fix(core): Count underscore-prefixed directories in token count tree

serhiizghama:
`--token-count-tree` silently dropped any directory whose name starts with an underscore — `__tests__`, `__mocks__`, `__generated__`, `_internal` and friends. The directory never showed up in the tree, and worse, its tokens weren't counted toward any ancestor's total, so the numbers were just wrong.

The cause is a namespace collision in the tree: directory names are stored as keys alongside the `_files` and `_tokenSum` metadata fields. To tell them apart, both the sum calculation and the reporter filtered out every key starting with `_`. But the value-type checks right next to it already exclude the metadata (`_files` is an array, `_tokenSum` is a number), so the `key.startsWith('_')` check was redundant — and it also matched real directories. Dropping it fixes both the totals and the rendering.

Example: with `src/__tests__/foo.test.ts` (100) and `src/index.ts` (50), `src/` now reports 150 tokens and `__tests__/` shows up, instead of `src/` showing 50 and the test dir vanishing.

Added regression coverage in both `buildTokenCountStructure` (sums roll up through the `__tests__` subtree) and the reporter (the directory is rendered with the right total). Both fail on `main` and pass here.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/cli/reporters/tokenCountTreeReporter.ts
+++ b/src/cli/reporters/tokenCountTreeReporter.ts
@@ -39,9 +39,11 @@ export const reportTokenCountTree = (
 };
 
 const displayNode = (node: TreeNode, prefix: string, isRoot: boolean, minTokenCount: number): void => {
-  // Get all directory entries (excluding _files and _tokenSum)
+  // Get all directory entries. The _files (array) and _tokenSum (number)
+  // metadata fields are excluded by the value-type check, so directories whose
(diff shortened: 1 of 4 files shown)


yamadashy (approved):
@serhiizghama 
Thanks for the PR!
I hadn't noticed this at all.
Nice catch, and thanks for the fix and the tests! 🙏

PR #1712, opened 2026-07-09
fix(core): split file paths on the POSIX separator so the tree doesn't collapse on Windows

serhiizghama:
By the time paths reach `sortPaths` and `generateFileTree` they're already normalized to forward slashes — globby returns `/`-paths on every OS, and `buildFileDisplayPath`/`toDisplayPath` explicitly rewrite separators to `/`. But both functions then split on the OS `path.sep`. On Windows that's a backslash, so a `/`-separated path never splits: `generateFileTree` puts every file at the root and the `<directory_structure>` section renders as a flat list of full paths instead of a tree, and `sortPaths` falls through to a plain whole-string `localeCompare` instead of the directory-aware ordering.

Fixed both to normalize to `/` before splitting, which is what `outputSplit.getRootEntry` already does for the same reason.

The existing sort tests didn't catch this because they build their inputs from `path.sep` too, so the input separator and the split separator always agree; the tree test used loose `toContain` checks that still pass on a collapsed tree. Added tests that feed a separator different from the host `path.sep` (backslashes on POSIX CI) to exercise the mismatch directly — both fail on `main` and pass with the fix.

### DIFF
--- a/src/core/file/filePathSort.ts
+++ b/src/core/file/filePathSort.ts
@@ -2,9 +2,16 @@ import path from 'node:path';
 
 // Sort paths for general use (not affected by git change count)
 // Uses decorate-sort-undecorate to pre-compute path.split() once per path
-// instead of O(N log N) repeated splits during comparisons
+// instead of O(N log N) repeated splits during comparisons.
+// Paths arrive normalized to POSIX separators (globby output, buildFileDisplayPath),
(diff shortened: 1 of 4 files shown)


yamadashy (approved):
@serhiizghama 
Nice catch on the Windows path split. Thanks for the fix and the focused tests! 🙏

PR #1725, opened 2026-07-13
fix(output): Widen Markdown fence so git diffs can't break it

serhiizghama:
The Markdown fence delimiter is computed only from file contents (`calculateMarkdownDelimiter`), but the template also wraps the directory structure and git diffs in a fixed triple-backtick fence. That's fine until a git diff contains backticks — which happens whenever you diff a Markdown file. A unified diff keeps the file's own fence lines as context, space-indented, and a space-indented bare triple backtick is a valid Markdown closing fence, so it ends the diff block early and the rest of the diff spills out as broken markup.

Repro with `--style markdown --include-diffs` on a repo with an uncommitted change near a code block in a Markdown file — the context line closes the diff block:

    ## Git Diffs Working Tree
    ```diff
    diff --git a/README.md b/README.md
    @@ -1,3 +1,3 @@
     ```        <- this closing fence in the diff ends the block early
    -old
    +new
    ```

The fix mirrors what's already done for file contents: fold the tree and diff text into the delimiter calculation so the fence is longer than any backtick run inside them, and use that delimiter for the directory-structure and git-diff fences instead of a fixed one. Added a regression test with a diff that carries a bare fence context line.

### DIFF
--- a/src/core/output/outputGenerate.ts
+++ b/src/core/output/outputGenerate.ts
@@ -52,9 +52,14 @@ const getCompiledTemplate = (style: string): Handlebars.TemplateDelegate => {
   return compiled;
 };
 
-const calculateMarkdownDelimiter = (files: ReadonlyArray<ProcessedFile>): string => {
-  const maxBackticks = files
-    .flatMap((file) => file.content.match(/`+/g) ?? [])
+// The Markdown template wraps file contents, the directory structure, and git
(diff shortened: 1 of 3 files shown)


yamadashy:
Thanks @serhiizghama! Nice catch, and the regression test is a good touch. Merging 🙏

yamadashy (approved):

PR #1726, opened 2026-07-14
fix(cli): Skip migrations for untrusted remote repos

ShiroKSH:
## Summary

- Skip legacy migration when the config pipeline is explicitly told not to load local repository config.
- Keep untrusted `--remote` clones from prompting for migration or rewriting legacy files.

## Root cause

`runRemoteAction` passed `skipLocalConfig: true`, but `buildMergedConfig` ran the migration before observing that flag.

## Tests

- `npm test -- --run` (1,523 passed, 14 skipped)
- `npm run lint`


yamadashy:
Thanks @ShiroKSH!

Good catch, migration should follow skipLocalConfig just like config loading does.
Merging 🙏

yamadashy (approved):

PR #1727, opened 2026-07-17
fix(output): Resolve Markdown language hint for extensionless filenames in subdirectories

serhiizghama:
In Markdown output the code-fence language hint is dropped for `Dockerfile`, `Makefile` and similar files whenever they sit in a subdirectory, while the same file at the repo root gets it right.

`getLanguageFromFilePath` builds the hint by splitting the path on `.` and looking up the last piece. The extension map has a few extensionless whole-filename keys (`dockerfile`, `makefile`, `cmake`), which only work when the path has no directory prefix — `docker/Dockerfile` has no dot at all, so `split(".")` keeps the whole path and the lookup misses:

```
getLanguageFromFilePath("Dockerfile")         // "dockerfile"
getLanguageFromFilePath("docker/Dockerfile")  // "" (should be "dockerfile")
getLanguageFromFilePath("src/Makefile")       // "" (should be "makefile")
```

So `docker/Dockerfile` renders as a bare fence with no language, but a root `Dockerfile` renders as ```` ```dockerfile ````. Fixed by matching against the basename before splitting on the dot. Files with normal extensions, dotfiles, and unknown types are unchanged.

Added regression tests for the subdirectory cases (plus a dotted directory segment and a Windows separator).

### DIFF
--- a/src/core/output/outputStyleUtils.ts
+++ b/src/core/output/outputStyleUtils.ts
@@ -219,7 +219,13 @@ const extensionToLanguageMap: Record<string, string> = {
  * @returns The language name for syntax highlighting, or empty string if unknown
  */
 export const getLanguageFromFilePath = (filePath: string): string => {
-  const extension = filePath.split('.').pop()?.toLowerCase();
+  // Match against the basename, not the whole path: the map holds a few
+  // extensionless whole-filename keys (dockerfile, makefile, cmake). A path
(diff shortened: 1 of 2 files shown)


yamadashy (approved):

yamadashy:
Thanks @serhiizghama!

Nice catch on the subdirectory basename, and good edge-case coverage. Merging 🙏

yamadashy (approved):

PR #1748, opened 2026-07-20
fix(core): normalize output file path to POSIX separators before ignoring

serhiizghama:
The output file is self-ignored by adding `path.relative(rootDir, outputPath)` to the ignore set in `getIgnorePatterns`. globby matches ignore patterns against forward-slash paths, but `path.relative` uses the OS separator — so on Windows a nested output path like `--output docs/out.xml` becomes `docs\\out.xml` and never matches, leaving the output file un-ignored and re-packed into itself on the next run. The default flat `repomix-output.xml` isn't affected since it has no separator.

Wrapped the relative path with the existing `toPosixPath` helper so the ignore pattern is always forward-slash, same normalization already used for the deferred ignore-control paths in this file. Added a test that mocks `path.relative` to return a backslash path and asserts the pattern comes out POSIX.

### DIFF
--- a/src/core/file/fileSearch.ts
+++ b/src/core/file/fileSearch.ts
@@ -425,7 +425,10 @@ export const getIgnorePatterns = async (rootDir: string, config: RepomixConfigMe
   // Add repomix output file
   if (config.output.filePath) {
     const absoluteOutputPath = path.resolve(config.cwd, config.output.filePath);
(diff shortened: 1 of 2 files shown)


yamadashy (approved):

yamadashy:
Thank you as always, @serhiizghama! Nice catch 🙏

PR #1749, opened 2026-07-21
fix(output): keep non-included files in --include-full-directory-structure tree

serhiizghama:
When `--include-full-directory-structure` is used together with `--include`, the Directory Structure section is supposed to show the entire repository tree, with only the included files having their contents packed. But non-included files were silently dropped from the tree.

The full-tree branch in `buildOutputGeneratorContext` correctly merges every repo file into `filePathsForTree` (included files plus everything else from `listFiles`), yet the tree was actually rendered from `filePathsByRoot`, which only carries the included files. Since `pack()` always passes `filePathsByRoot` for the default `target-relative` style, the merged `filePathsForTree` was dead code in that path and the extra files never made it into the output.

Repro — a repo with `README.md`, `src/index.ts`, `src/b/other.ts`, `docs/guide/intro.md`:

```
repomix --include "src/**/*.ts" --include-full-directory-structure
```

Before, the tree listed the `docs/` and `docs/guide/` directories but not `README.md` or `docs/guide/intro.md` — only the included `.ts` files. Now the whole tree shows up, and just the `src/**/*.ts` files get content blocks.

The fix renders the already-merged `filePathsForTree`/`directoryPathsForTree` directly in full-tree mode instead of the included-only `filePathsByRoot`. The existing root-label test only asserted directories and the included file, so it missed this — I extended coverage to assert the non-included files appear too (it fails without the fix).

### DIFF
--- a/src/core/output/outputGenerate.ts
+++ b/src/core/output/outputGenerate.ts
@@ -425,7 +425,13 @@ export const buildOutputGeneratorContext = async (
   // Generate tree string - use multi-root format if filePathsByRoot is provided
   // generateTreeStringWithRoots handles single root case internally
   let treeString: string;
-  if (filePathsByRoot) {
+  if (shouldUseFullTree) {
+    // Full-tree mode already merged every repo file (included + the rest) and every
+    // directory into filePathsForTree/directoryPathsForTree above. Render those directly:
+    // filePathsByRoot only carries the included files, so it would drop the non-included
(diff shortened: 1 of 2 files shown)


yamadashy (approved):

yamadashy:
@serhiizghama Merged, will be in the next release. Always appreciate your contributions 🙏

PR #1753, opened 2026-07-24
feat(mcp): Add --sandbox to confine the MCP server to a directory

huy-trn:
## Discussion
https://github.com/yamadashy/repomix/issues/1754

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/mcp/tools/packCodebaseTool.ts
+++ b/src/mcp/tools/packCodebaseTool.ts
@@ -1,20 +1,45 @@
+import fs from 'node:fs/promises';
 import path from 'node:path';
 import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import type { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+import { braceExpand } from 'minimatch';
 import { z } from 'zod';
 import { runCli } from '../../cli/cliRun.js';
 import type { CliOptions } from '../../cli/types.js';
 import { defaultFilePathMap } from '../../config/configSchema.js';
+import { logger } from '../../shared/logger.js';
+import { splitPatterns } from '../../shared/patternUtils.js';
+import type { McpServerConfig } from '../mcpServer.js';
+import { isEscapingPath, resolveWithinRoot } from '../pathScope.js';
 import {
   buildMcpToolErrorResponse,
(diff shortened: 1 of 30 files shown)


src/mcp/tools/packCodebaseTool.ts
```diff
@@ -101,9 +166,26 @@ export const registerPackCodebaseTool = (mcpServer: McpServer) => {
...
           topFilesLen: topFilesLength,
           quiet: true,
+          // Untrusted-workspace lockdown — one set, all gated on --sandbox:
+          //  - skip EVERY config file: the workspace's repomix.config.* AND the
+          //    operator's global one, since either could set
+          //    output.instructionFilePath to read an out-of-workspace file into the
+          //    agent-visible output, or input.processors to run commands;
+          //  - confine the file search to the root, a syntax-agnostic backstop behind
+          //    the include/ignore pattern guard.
+          ...(config.sandboxed ? { skipLocalConfig: true, skipGlobalConfig: true, confineToBaseDir: true } : {}),
```
yamadashy: @huy-trn Sandbox packs still run `git -C <workspace> log` here (sortByChanges is on by default), which reads the untrusted workspace's `.git/config` (e.g. `gpg.program` via `log.showSignature`). Disable it in the lockdown:

```suggestion
          ...(config.sandboxed
            ? { skipLocalConfig: true, skipGlobalConfig: true, confineToBaseDir: true, gitSortByChanges: false }
            : {}),
```

huy-trn: Good catch, thanks. Fixed in https://github.com/yamadashy/repomix/pull/1753/commits/d08872f34d0a3e096e6bcab792c1e743bb9f56a7 — added `gitSortByChanges: false` to the sandbox lockdown as suggested, so `sortByChanges` no longer runs `git -C <workspace> log` against the workspace's `.git/config`. Also rebased on latest `main` and cleared the `alikes` typo that was failing `Check typos`.

yamadashy (approved):
Thanks @huy-trn — great to have you back, and this is clean, well-tested work!

Merging.

I'll follow up on the `--sandbox-strict` direction in #1754 soon.

PR #1772, opened 2026-08-02
fix(core): accept explicit remote URLs whose repo name starts with a dot (e.g. .github)

serhiizghama:
`parseRemoteValue` rejects some perfectly valid repositories with `Invalid remote repository URL or repository shorthand (owner/repo)`. The clearest example is an org profile repo — `https://github.com/microsoft/.github` throws, even though the repo exists and clones fine.

The cause is the owner/repo re-check after `git-url-parse` has already parsed the URL: it runs the parsed `owner/repo` back through `isValidShorthand`, whose regex forces every name segment to start and end with an alphanumeric. That rule is right for its real job — telling an ambiguous `owner/repo` shorthand apart from a local path — but it's wrong to reuse it on an explicit URL, since hosts happily allow names that lead or trail with `.`/`-`/`_` (`.github` being the obvious one). The two parsers already disagree here: `parseGitHubRepoInfo` extracts `{owner: "microsoft", repo: ".github"}` from the same input without complaint.

For public GitHub the archive fast-path hides this, but the clone path (non-GitHub hosts, or the archive→clone fallback) hits the throw and reports a misleading "invalid owner/repo" for a valid repo. `isValidRemoteValue` returns `false` for it too.

The fix only skips the shorthand re-check when the input is an explicit URL (`isExplicitRemoteUrl`); ambiguous shorthand still goes through the strict regex, so `user/.repo` and friends stay rejected as before. Added tests for the dotfile-name URL/SSH cases and a guard that shorthand `user/.repo` still throws.

### DIFF
--- a/src/core/git/gitRemoteParse.ts
+++ b/src/core/git/gitRemoteParse.ts
@@ -2,7 +2,7 @@ import gitUrlParse, { type GitUrl } from 'git-url-parse';
 import { RepomixError } from '../../shared/errorHandle.js';
 import { logger } from '../../shared/logger.js';
 import { assertNotMetadataEndpoint } from './gitMetadataEndpoint.js';
-import { isExplicitRemoteUrl, isValidShorthand } from './gitRemoteUrl.js';
+import { isExplicitRemoteUrl, isValidExplicitUrlOwnerRepo, isValidShorthand } from './gitRemoteUrl.js';
 
 interface IGitUrl extends GitUrl {
   commit: string | undefined;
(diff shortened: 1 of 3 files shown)


serhiizghama:
Good catch on the doubled-slash case — `isValidExplicitUrlOwnerRepo` now validates explicit-URL segments (non-empty, still allowing leading/trailing `.`/`-`/`_`) instead of skipping validation entirely for explicit URLs. Added a regression test for `user//repo`.

One divergence from the suggestion: didn't reject `user/repo/extra` as an extra path component. Checked what `git-url-parse` actually produces for it — `owner: 'user/repo', name: 'extra'` — structurally identical to the GitLab nested-group paths this file already supports and tests (e.g. `some/path/username/repo`, covered by the existing 'should get correct branch name from url' test). The `slice(-2)` logic is deliberately host-agnostic to support arbitrary-depth subgroups, so treating a 3rd segment as invalid would break that. Rejecting it specifically for github.com would need host-aware logic this file doesn't have today.

Also added coverage for the trailing/leading `.`/`-`/`_` boundary cases from your nitpick on the test file.

144/144 in `tests/core/git/`, tsc/biome clean.

serhiizghama:
Rebased onto main to clear the conflict with the metadata-endpoint security fix that landed since — no logic changes, just resolving the import/guard overlap. 180/180 green in tests/core/git/, lint clean.

serhiizghama:
The one nitpick left open here was handled back in 1b36fed9: `should accept explicit URLs whose owner/repo name leads or trails with . - _` covers names ending in `.`, `-` and `_` as well as names beginning with `-` and `_`, and there's a separate case pinning the `user//repo` rejection. Auto-incremental review is off in this repo, so the review decision never got refreshed after that commit and the PR still shows as changes-requested.

@coderabbitai review

yamadashy (approved):
@serhiizghama Thanks, nice fix! Will be in the next release 🙏

PR #1773, opened 2026-08-03
fix(file): handle escaped gitignore device-name rules

### LINKED ISSUE #1765: Repomix now fails on repos using backslash in .gitignore patterns
## Issue Description

When running `repomix` on repositories that contain `.gitignore` files with certain backslash-escaped patterns, it fails with error during the file filtering phase.

The underlying path validation (likely from the `ignore` npm package) throws an error because it encounters a string it doesn't consider a valid relative path.

## Steps to Reproduce

You can reproduce this by running Repomix against the following repositories:

**Test Case 1: [Microsoft Data Formulator](https://github.com/microsoft/data-formulator)**

1. Run `repomix --remote https://github.com/microsoft/data-formulator`
    - **Error:** `Error filtering files: path should be a path.relative()d string, but got "\.\NUL"`
    - **Failing file:** [data-formulator/.gitignore](https://github.com/microsoft/data-formulator/blob/main/.gitignore)

**Test Case 2: [Claude Code (Best)](https://github.com/claude-code-best/claude-code)**

1. Run `repomix --remote https://github.com/claude-code-best/claude-code`
    - **Error:** `Error filtering files: path should be a `path.relative()`d string, but got "\*.pid.lock"`
    - **Failing file:** [claude-code/packages/cloud-artifacts/.gitignore](https://github.com/claude-code-best/claude-code/blob/main/packages/cloud-artifacts/.gitignore)

## Solution Discussion

To fix this robustly, we need to look at both the Git specification and how the file-filtering library processes paths.

### 1. What's the real meaning of a backslash in a `.gitignore` definition?

According to the [official Git documentation](https://git-scm.com/docs/gitignore):

* A backslash (`\`) is strictly used as an **escape character**. It is used to escape characters that would otherwise be treated as special syntax (e.g., `\!`, `\#`, `\*`, `\?`).
* To ignore a literal backslash, it must be escaped as well (`\\`).
* **Crucially:** A backslash is **never** used as a directory separator in `.gitignore`, even on Windows. Directory separators are always forward slashes (`/`).

### 2. How to handle this best in Repomix?

Currently, it seems Repomix (or its dependency, like `ignore`) is either incorrectly parsing these escaped rules as file paths, or failing when testing actual files/strings that contain backslashes.

To respect the `.gitignore` meaning while preventing failing with error, Repomix could implement the following:

* **Sanitize Paths Before Validation:** Ensure that any string passed to the path-checking/filtering utility is properly normalized to a valid relative path (using forward slashes `/`).
* **Graceful Degradation (Try/Catch):** Wrap the file filtering check (e.g., the `.ignores()` call) in a `try...catch` block. If a specific path or rule causes the underlying library to throw a `path should be a path.relative()d string` error, Repomix should log a verbose warning and safely skip that specific file/rule, rather than failing the entire packing process with error.
* **Respect Escape Sequences:** Ensure the internal globber or gitignore parser isn't accidentally transforming `\*` or `\.` into malformed file paths during its traversal phase.

By implementing a fallback/catch for malformed paths, Repomix will become much more resilient to the wild variations of `.gitignore` files found in the open-source ecosystem.

### Usage Context

Repomix CLI

### Repomix Version

1.17.0

### Node.js Version

v26.5.0

**EDIT**: Replace AI Agent `crash` wording. There's no crash, just fail with error. Also fixed links

JD114514-CPU:
## Summary
- fall back to manual `.gitignore` replay when `globby` crashes on escaped Windows device-name rules like `\\.\\NUL`
- reuse that fallback for `searchFiles`, `listFiles`, and `listDirectories`
- add a regression test covering an escaped device-name rule plus a normal ignore rule in the same file

## Validation
- `npm run test -- tests/core/file/fileSearch.gitignoreSpec.test.ts`
- `npm run test -- tests/core/file/fileSearch.test.ts`
- `npm run build`
- `node --enable-source-maps --trace-warnings bin/repomix.cjs --remote https://github.com/microsoft/data-formulator --output E:\CodexGitHubScout\tmp\repomix-data-formulator.xml`

Closes #1765.

### DIFF
--- a/src/core/file/fileSearch.ts
+++ b/src/core/file/fileSearch.ts
@@ -22,6 +22,7 @@ export interface FileSearchResult {
 // descriptor exhaustion that unbounded `Promise.all` could cause.
 const EMPTY_DIR_CHECK_CONCURRENCY = 20;
 const IGNORE_CONTROL_FILE_NAMES = new Set(['.gitignore', '.ignore', '.repomixignore']);
+const GITIGNORE_RELATIVE_PATH_ERROR_FRAGMENT = 'path should be a `path.relative()`d string';
 
 // No per-directory ignore-pattern check is needed here. The `directories` array
 // comes from globby with the same `ignore` patterns (e.g. `dist/**`), which
@@ -90,6 +91,11 @@ export const normalizeGlobPattern = (pattern: string): string => {
 
 const toPosixPath = (value: string): string => value.replace(/\\/g, '/');
 
+type ManualGitignoreRule = {
+  negative: boolean;
+  pattern: string;
+};
+
 // Canonical posix form of a deferred ignore pattern: forward slashes and no
 // trailing slash. Detection (isIgnoreControlFilePattern) and post-filtering
 // (filterDeferredIgnoredFiles) must share this so a pattern that is deferred is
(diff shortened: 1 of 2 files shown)


yamadashy:
@JD114514-CPU Thanks for looking into #1765. Closing this one, though:

- The diff converts the whole file to CRLF line endings, so every line shows as rewritten and the change can't be meaningfully reviewed.
- Detecting the failure by matching an error message string and falling back to a manual reimplementation of gitignore matching isn't a direction I want to take — gitignore semantics are subtle and this would diverge from Git's actual behavior. The root cause looks like an upstream issue in globby's gitignore handling on Windows, so I'd rather fix it there or ship a minimal targeted mitigation.

The bug itself is tracked in #1765.

PR #1775, opened 2026-08-04
test(core): Add regression tests for the Windows .gitignore backslash crash

### LINKED ISSUE #1765: Repomix now fails on repos using backslash in .gitignore patterns

yamadashy:
## Summary

Fixes #1765. On Windows, packing failed with ``path should be a `path.relative()`d string, but got "\.\NUL"`` when a target repo's .gitignore mixes a backslash-escaped rule with a negation (`!`) rule.

The bug was in globby, not repomix, and is **already fixed upstream** in [globby 16.2.3](https://github.com/sindresorhus/globby/issues/280), which #1803 just brought in via renovate. This PR is the regression tests only, so we notice if it ever comes back.

## Root cause

globby **16.2.2**'s new gitignore pruning (`createNameComparer`) fed raw .gitignore rule text into the `ignore` package's `.ignores()` as if it were a path. Only `ignore`'s win32 branch rewrites `\.\NUL` to `/./NUL`, which then fails its relative-path validation, hence Windows-only. The comparison only runs when negation rules are present, so minimal fixtures passed everywhere and only a verbatim replay of the 419-line [data-formulator .gitignore](https://github.com/microsoft/data-formulator/blob/main/.gitignore) reproduced it.

## Changes

- 8 regression tests (`tests/core/file/fileSearch.gitignoreBackslash.test.ts`), including the verbatim data-formulator .gitignore, each with and without a git root
- Exclude `tests/fixtures/gitignore/` from the typos check, since the fixture must stay byte-identical to the real-world file

## Test plan

- [x] Windows CI legs (Node 22/24/26 + Bun) reproduced the exact reported error on globby 16.2.2, and pass on 16.2.3
- [x] `npm run test` / `npm run lint`


yamadashy:
Note to self: HEAD currently pins globby 16.2.1 + ignore 7.0.5 (the v1.17.0 dependency set) purely to replay the reported environment on the Windows CI leg. Reverting before this leaves draft.

yamadashy:
Upstream fixed in globby 16.2.3 (sindresorhus/globby#280 — the failing test from this investigation was adopted into the fix). The 16.2.1 pin stays for now since `.npmrc` `min-release-age=7` blocks a 1-day-old release; renovate will propose the 16.2.3 bump once the cooldown passes, and this PR's regression tests will verify it on Windows CI.

PR #1799, opened 2026-08-11
fix(security): Neutralize untrusted repo git config in log and diff subprocesses

yamadashy:
## Summary

Packing runs `git log` inside the target directory by default (`output.git.sortByChanges`), and `git diff` when `output.git.includeDiffs`/`includeLogs` are on. git honors a repository's own `.git/config` while merely reading history, so when the packed directory is not one the user created (a downloaded checkout, or an agent packing an arbitrary repo), that config can name executables git will run:

- `gpg.program` via `log.showSignature` — reported in GHSA-4p5g, reproduced end to end against `repomix` 1.18.0 (a plain `repomix` in a crafted directory wrote the payload marker with the user's privileges)
- `diff.external`, per-attribute `textconv` drivers, and `core.fsmonitor` — found firing the same way while scoping the fix

This is the same class of bug already closed for the website ZIP upload and the sandboxed MCP path; this brings the CLI's own git calls in line.

## Fix

Every git command that runs inside the target directory is hardened:

- log: `-c core.fsmonitor=` + `--no-show-signature`
- diff: `-c core.fsmonitor=` + `--no-ext-diff --no-textconv`

Command flags disable the log/diff drivers; `core.fsmonitor` has no flag so it is overridden to empty.

## Not a breaking change

Verified against git 2.53 that the hardened commands produce **byte-for-byte identical output on a legitimate repository** (both `log --name-only` and `diff`). `sortByChanges` stays on by default; nothing about normal use changes.

Flipping `sortByChanges` to default-off was considered and rejected: it degrades output quality for everyone, is itself a breaking change, and an attacker's own `repomix.config` could turn it back on — so it would not actually close the vector.

## Tests

- Updated the mocked assertions to pin the exact hardened argument lists.
- Added an end-to-end test (nothing mocked): a real repository carrying a malicious `.git/config` is pushed through `execGitLogFilenames`, asserting the payload does **not** run, with a positive control proving the payload is live on the host. Removing the hardening turns this test red.
- Full suite green: 1800 tests.

## Note on the report

The advisory claims `sortByChanges` became default-`true` only in 1.18.0; it has actually been default-`true` since v0.2.41 (2025-03-16). The report appears to have read the CLI-option schema (no default) rather than the config schema. This is a long-standing default, not a 1.18.0 regression.

### DIFF
--- a/src/core/git/gitCommand.ts
+++ b/src/core/git/gitCommand.ts
@@ -8,6 +8,20 @@ import { redactErrorMessage, redactUrl } from '../../shared/urlRedact.js';
 
 const execFileAsync = promisify(execFile);
 
+// A repository's own .git/config can name executables that git will run while
+// merely reading history: gpg.program (reached through log.showSignature),
+// diff.external, per-attribute textconv drivers, and core.fsmonitor. repomix
+// runs these read-only git commands inside directories it did not create — a
+// local pack of an untrusted checkout, or any working directory — so a malicious
+// repository could otherwise turn `git log`/`git diff` into command execution on
+// this host. Every such invocation is neutralized: the command flags disable the
+// log/diff drivers and core.fsmonitor (which has no flag) is overridden to empty.
+// Verified against git 2.53 to block all four vectors while leaving output for a
+// legitimate repository byte-for-byte unchanged.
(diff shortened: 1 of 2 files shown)


yamadashy:
Closing: this fixes an unpublished advisory (GHSA-4p5g) and should not be developed in the open. Moving to the advisory's temporary private fork.

PR #1808, opened 2026-08-18
fix(cli): Reject split sizes below one byte

Whxuan0701:
## Summary

- Reject positive split-size values that floor to zero bytes after unit conversion.
- Return a clear minimum-size error instead of allowing a zero-byte threshold.
- Add a regression test for a sub-byte kilobyte value.

## Validation

- [x] Run `npm run test` (152 files, 1807 tests)
- [x] Run `npm run lint`

## Risks

- Low. Valid sizes of at least one byte are unchanged.


yamadashy (approved):
@Whxuan0701 Thanks for the fix! Will be in the next release 🙏

PR #1809, opened 2026-08-18
fix(cli): Respect directory boundaries when classifying paths

Whxuan0701:
## Summary

- Use `path.relative()` results to determine whether output and skill paths are inside their base directories.
- Keep sibling directories with matching string prefixes outside the base directory.
- Add regression tests for CLI display paths and personal skill location classification.

## Validation

- [x] Run `npm run test` (152 files, 1808 tests)
- [x] Run `npm run lint`

## Risks

- Low. Paths genuinely inside the configured base directory keep their existing behavior.


yamadashy (approved):

yamadashy (approved):
@Whxuan0701 Thanks, good catch on the prefix check! Will be in the next release 🙏

PR #1810, opened 2026-08-18
fix(skill): Derive repository names from GitHub URLs

Whxuan0701:
## Summary

- Parse standard GitHub URLs by owner and repository before deriving remote skill names.
- Preserve the existing shorthand and SSH URL fallbacks.
- Add regression coverage for GitHub branch and commit URLs.

## Validation

- [x] Run `npm run test` (152 files, 1808 tests)
- [x] Run `npm run lint`

## Risks

- Low. The URL-specific path only applies to `github.com`; other remote formats continue through the existing fallback logic.

### DIFF
--- a/src/core/skill/skillUtils.ts
+++ b/src/core/skill/skillUtils.ts
@@ -124,6 +124,18 @@ export const extractRepoName = (url: string): string => {
     cleanUrl = cleanUrl.slice(0, -4);
   }
 
+  try {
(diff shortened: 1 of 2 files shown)


yamadashy (approved):

yamadashy (approved):
@Whxuan0701 Thanks for the fix! Will be in the next release 🙏

PR #1820, opened 2026-08-27
fix(core): Preserve data URI parameters when truncating base64

### LINKED ISSUE #1819: truncateBase64Content silently drops data-URI parameters due to repeated capture group
## Summary

## Location

- Relevant code:

```ts
  'g',
);
...
});
```

## Problem

## Trigger / Reproduction

Based on source reading:

## Expected Behavior

## Actual Behavior

## Impact

## Suggested Direction

## Evidence

cyphercodes:
Fixes #1819

## Summary

- Preserve the full data URI parameter run before `;base64,` when truncating long base64 payloads.
- Add regression coverage for data URIs with multiple parameters.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/src/core/file/truncateBase64.ts
+++ b/src/core/file/truncateBase64.ts
@@ -7,7 +7,7 @@ const MIN_CHAR_TYPE_COUNT = 3;
 
 // Pre-compiled regex patterns (avoid re-creation per file)
 const dataUriPattern = new RegExp(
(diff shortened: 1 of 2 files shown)


fuleinist:
Resolved by #1821 (duplicate fix). Thank you @cyphercodes!

yamadashy (approved):
@cyphercodes Thanks for the fix! Nice catch. Will be in the next release 🙏

yamadashy (approved):

PR #1821, opened 2026-08-28
fix(truncateBase64): preserve all parameters in data URIs with multiple params

### LINKED ISSUE #1819: truncateBase64Content silently drops data-URI parameters due to repeated capture group

fuleinist:
## Summary

Fixes #1819

The `truncateBase64Content` function was silently dropping data URI parameters when multiple parameters were present. Only the last parameter in a sequence was being preserved in the output.

## Root Cause

The regex pattern used a repeated capturing group `(;\[a-zA-Z0-9\-=\]+)*` which only captures the final iteration. When multiple parameters like `charset=utf-8;name=file.png` were present, only the last one survived.

## Fix

Changed the pattern from a repeated capturing group to a non-capturing group wrapped in a capturing group: `((?:;\[a-zA-Z0-9\-=\]+)*)`. This captures all parameters as a single group, preserving them in the output.

## Changes

- **src/core/file/truncateBase64.ts**: Fixed the regex pattern on line 10 (1 line changed)
- **tests/core/file/truncateBase64.test.ts**: Added 2 regression test cases

## Testing

All 19 tests pass (17 existing + 2 new). New tests verify:
- Data URIs with 2 parameters preserve both
- Data URIs with 3+ parameters preserve all

### DIFF
--- a/src/core/file/truncateBase64.ts
+++ b/src/core/file/truncateBase64.ts
@@ -7,7 +7,7 @@ const MIN_CHAR_TYPE_COUNT = 3;
 
 // Pre-compiled regex patterns (avoid re-creation per file)
 const dataUriPattern = new RegExp(
-  `data:([a-zA-Z0-9\\/\\-\\+]+)(;[a-zA-Z0-9\\-=]+)*;base64,([A-Za-z0-9+/=]{${MIN_BASE64_LENGTH_DATA_URI},})`,
+  `data:([a-zA-Z0-9\\/\\-\\+]+)((?:;[a-zA-Z0-9\\-=]+)*);base64,([A-Za-z0-9+/=]{${MIN_BASE64_LENGTH_DATA_URI},})`,
   'g',
 );
 const standaloneBase64Pattern = new RegExp(`([A-Za-z0-9+/]{${MIN_BASE64_LENGTH_STANDALONE},}={0,2})`, 'g');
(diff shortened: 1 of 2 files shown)


fuleinist:
Addressed both CodeRabbit findings in commit 91882402:

1. Major (regex): added dot to param value char class [a-zA-Z0-9._\-=]+ in dataUriPattern
2. Minor (tests): strengthened assertions to use toBe(expected) with full expected strings

Regression tests cover 2-param and 3-param data URIs with dots in param values. Re-requesting review.

fuleinist:
@coderabbitai re-review

Both findings from the CHANGES_REQUESTED review are addressed in `91882402` (current HEAD):

1. **Major — param value grammar:** `dataUriPattern` now accepts dots in parameter values (`[a-zA-Z0-9._\-=]+`), so `name=icon.png` / `name=logo.svg` match and truncate correctly instead of passing through unchanged.
2. **Minor — assertion strength:** regression tests assert the complete truncated URI via `toBe(expected)` (prefix through `;base64,` + first 32 payload chars + `...`), not `toContain`.

Re-verified locally on `91882402`: `vitest run tests/core/file/truncateBase64.test.ts` → 19/19 pass, `tsc --noEmit` clean, CI green.

yamadashy:
@fuleinist Closing as #1819 was fixed by #1820, which landed first with the same regex change. Thanks for the PR, and I'd be glad to see more from you!

PR #1827, opened 2026-08-29
fix(security): Keep dependency code out of the npm-publishing job

yamadashy:
A security review of the release workflows found that `npm-publish.yml` holds `id-token: write` while running ~400 third-party packages.

## Problem

`id-token: write` puts `ACTIONS_ID_TOKEN_REQUEST_URL` / `ACTIONS_ID_TOKEN_REQUEST_TOKEN` in the job environment, and any code running there can exchange them for an npm publish token. The single job held that permission while running:

- `npm install -g npm@latest` (unpinned, floating)
- `npm ci` (~400 packages, lifecycle scripts enabled)
- lint, the test suite, and the build

all **before** `npm publish`. One compromised dev dependency could have published a malicious `repomix` release carrying valid provenance.

## Change

Split into three jobs in the same file:

- **`check`** — `contents: read` only. Runs lint and the test suite, which is where the whole dependency tree actually executes (vitest, the linters' native binaries, everything `src/` imports). Never touches the tarball.
- **`build`** — `contents: read` only. Runs `npm ci` and `npm pack`, nothing else. The code that can reach the package contents is reduced to npm itself, the install scripts (only `esbuild` on linux), and `tsc` via `prepare`. Dev-only dependencies have no other path to end users, so keeping them out of this job is what narrows the build-time tampering risk rather than only the token-minting risk.
- **`publish`** — `needs: [build, check]`, holds `id-token: write`, and runs nothing but `npm publish <tgz> --provenance`. No checkout, no `npm ci`, no global npm install. `check` in `needs` is the only gate on lint/tests now that `build` does not run them.

Also drops `npm install -g npm@latest`: Node 24.15.0 bundles npm 11.12.1, past the 11.5.1 trusted-publishing floor, and a floating global install inside the credentialed job was itself worth removing.

## Verification

Checked against the npm CLI source rather than assumed, since a mistake here breaks releases:

- **Provenance needs no checkout.** `libnpmpublish/lib/provenance.js` builds the attestation entirely from GitHub Actions env vars (`GITHUB_WORKFLOW_REF`, `GITHUB_SHA`, ...); it never reads `.git`. The only `readFile` in that module belongs to `--provenance-file`, which is not used here.
- **Tarball contents are unchanged.** `npm publish <tgz>` streams the file verbatim (`pacote/lib/file.js`), and the digest is computed from the manifest, not the CLI argument. `files` and `.npmignore` are resolved at pack time in the build job exactly as they are today.
- **Lifecycle scripts are preserved.** The only relevant script is `prepare: npm run build`, which now runs during `npm pack` in the build job. Summed across the two jobs the set is identical to today; the publish job runs none.
- **Trusted publishing still matches.** The npm-side config keys on the workflow *filename*, and `GITHUB_WORKFLOW_REF` is per-file rather than per-job, so keeping both jobs in `npm-publish.yml` needs no change on npm. (Moving `publish` into a reusable workflow *would* break it, so that was avoided.)

## Known cosmetic change

`gitHead` disappears from the registry manifest — it is populated only on the directory publish path. The commit is still bound, more strongly, by the provenance attestation's `resolvedDependencies[0].digest.gitCommit`. `_npmVersion` also goes from `12.0.2` to `11.12.1`.

## Not included

Binding the job to a GitHub environment was also recommended. That needs a repo settings change plus a matching entry in the npm trusted-publisher config, and referencing an environment that does not exist yet auto-creates it with no protection rules. It should be done deliberately rather than as a side effect of this PR.

## Checklist

- [x] Run `npm run test`
- [x] Run `npm run lint`

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -6,13 +6,16 @@ name: npm-publish
 #   tree (vitest, the linters' native binaries, everything `src/` imports),
 #   so it holds no publish permission and never touches the tarball.
 # - build produces the tarball. It only runs `npm ci` and `npm pack`, so the
-#   code that can reach the package contents is limited to npm itself, the
-#   install scripts, and the TypeScript compiler. Dev-only dependencies have
-#   no other path to end users, which is why they are kept out of this job.
+#   code that executes there is limited to npm itself, the install scripts,
+#   and `prepare` (rimraf + tsc, run by both commands). The full dependency
+#   tree is still installed, but no other dev-only dependency is executed,
+#   and executing here is their only path to the package contents.
 # - publish holds `id-token: write`. That permission puts the OIDC minting
 #   credentials in the job environment, and anything running there can
-#   exchange them for an npm publish token, so this job runs nothing but
-#   `npm publish` on the tarball the build job handed over.
+#   exchange them for an npm publish token. This job therefore has no
+#   checkout and no `npm ci`: apart from the artifact download and Node setup
+#   actions, it only verifies the tarball digest and the npm version, then
+#   runs `npm publish` on the tarball.
 
 on:
   workflow_dispatch:
@@ -45,8 +48,12 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # `lint` runs biome and oxlint in write mode. The build job packs the
+      # committed tree from its own checkout, so any auto-fix here would mean
+      # the tested code and the published code differ: fail instead, as ci.yml
+      # does.
       - name: Lint
-        run: node --run lint
+        run: node --run lint && git diff --exit-code
 
       - name: Test
         run: node --run test-coverage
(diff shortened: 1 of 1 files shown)


yamadashy:
Addressed in `f544ea82`. All four were valid.

- **Tarball hand-off unverified** — the build job now records the sha256 as a job output and the publish job checks it after download. Used `--strict`, which also rejects a malformed line, so a lost or empty job output fails closed rather than checking nothing (verified against GNU coreutils, including the empty-digest case).
- **`.tool-versions` parsing** — switched to `awk '$1 == "nodejs" { print $2; exit }'`, so a fallback version list or a trailing comment cannot produce a multi-line `GITHUB_OUTPUT` value.
- **npm floor only in a comment** — added the explicit `sort -V -C` check in the publish job. Confirmed it passes at 11.12.1 and rejects 11.4.0.
- **Unused `contents: read`** — dropped from the publish job.

On the dry-run dispatch: agreed that is the right rehearsal, and it is worth doing on this branch before the next release since `workflow_dispatch` can target it directly.

🤖

PR #1837, opened 2026-09-05
fix(truncateBase64): keep every data-URI parameter, not just the last

### LINKED ISSUE #1819: truncateBase64Content silently drops data-URI parameters due to repeated capture group

drexthealpha:
Fixes #1819

**Problem.** `dataUriPattern` captures parameters with a repeated capturing group, `(;[a-zA-Z0-9\-=]+)*`. JavaScript keeps only the last iteration of a repeated group, so a data URI with two or more parameters loses all but the final one when the match is rebuilt in the replace callback. Truncation corrupts the URI rather than only shortening its base64 payload.

**Change.** One character class: the inner group becomes non-capturing and the whole run is captured instead, `((?:;[a-zA-Z0-9\-=]+)*)`. `params` then holds every parameter. Zero- and one-parameter URIs match and render exactly as before, which is why the existing `charset` test did not catch this.

**Verify.**

```
npx vitest run tests/core/file/truncateBase64.test.ts
```

18 pass. The added test fails on `main` and passes with the fix:

```
input   data:image/svg+xml;charset=utf-8;foo=bar;base64,PHN2...
before  data:image/svg+xml;foo=bar;base64,PHN2...        <- charset dropped
after   data:image/svg+xml;charset=utf-8;foo=bar;base64,PHN2...
```

`npx biome check` is clean on both files.

### DIFF
--- a/src/core/file/truncateBase64.ts
+++ b/src/core/file/truncateBase64.ts
@@ -7,7 +7,7 @@ const MIN_CHAR_TYPE_COUNT = 3;
 
 // Pre-compiled regex patterns (avoid re-creation per file)
 const dataUriPattern = new RegExp(
-  `data:([a-zA-Z0-9\\/\\-\\+]+)(;[a-zA-Z0-9\\-=]+)*;base64,([A-Za-z0-9+/=]{${MIN_BASE64_LENGTH_DATA_URI},})`,
+  `data:([a-zA-Z0-9\\/\\-\\+]+)((?:;[a-zA-Z0-9\\-=]+)*);base64,([A-Za-z0-9+/=]{${MIN_BASE64_LENGTH_DATA_URI},})`,
   'g',
 );
 const standaloneBase64Pattern = new RegExp(`([A-Za-z0-9+/]{${MIN_BASE64_LENGTH_STANDALONE},}={0,2})`, 'g');
--- a/src/core/file/truncateBase64.ts
+++ b/src/core/file/truncateBase64.ts
@@ -6,8 +6,15 @@ const MIN_CHAR_DIVERSITY = 10;
 const MIN_CHAR_TYPE_COUNT = 3;
 
 // Pre-compiled regex patterns (avoid re-creation per file)
+// Parameter names and values are RFC 2045 tokens, and RFC 2397 allows a value to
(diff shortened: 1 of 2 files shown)


drexthealpha:
On the parameter-grammar suggestion: that limitation is real but pre-existing, and this PR does not change it either way. Checked both patterns against `data:text/plain;foo=bar.baz;base64,...`:

```
main     matches dotted param: false
this PR  matches dotted param: false
```

A parameter containing `.` or a percent-encoded value fails to match on `main` too, so such a URI is left untruncated rather than corrupted — a smaller problem than the one #1819 reports, and a different one. Widening the character class also widens what counts as a data URI, which deserves its own issue and its own test cases rather than riding along here.

Happy to open that as a follow-up if you would like it, but I have kept this PR to the repeated-capture-group bug so it stays reviewable in a minute.

drexthealpha:
Addressed the review comment.

The parameter portion matched only `[a-zA-Z0-9\-=]`, so an ordinary `;name=photo.v2.png` or a percent-encoded `;title=a%20b` made the whole data-URI pattern miss and the payload was left untruncated — the same false negative this PR is about, one level down.

Parameter names and values are RFC 2045 tokens and RFC 2397 allows the value to be percent-encoded, so the class is now that set, pulled out as `DATA_URI_TOKEN` so both halves stay in step. It still excludes `;` and `,` so a parameter cannot run past its own separator, and the base64 payload-length behaviour is unchanged.

The regression case covers punctuation and percent-encoding in one URI:

```
data:image/svg+xml;charset=utf-8;name=photo.v2.png;title=a%20b;base64,...
```

I stashed the source and watched that case go red on the old pattern before keeping it, so it is testing the fix rather than describing it. 19 tests in the file pass, `biome check` is clean and `tsc --noEmit` exits 0.

yamadashy:
@drexthealpha Closing as #1819 was fixed by #1820, which landed first. If the wider RFC 2045 parameter grammar turns out to matter in practice, happy to look at it as a separate issue. Thanks for the PR, and I'd be glad to see more from you!

